PR #5, opened 2018-03-01
fix Tree.get() to correctly return data

jkrysl:
There was a bug where Tree.get() failed with 'KeyError: None'. Inverting the logic works as intended, with no name given return the whole dict, otherwise return only the atribute with name given.


psss:
Nice catch! Thanks for the fix.

PR #6, opened 2018-03-06
ability to add to parent atribute at child using '+' at the end

jkrysl:
This code allows user to add value to inherited attribute.
If there is inherited attribute, it converts it to a list and appends the new value.

Example:
parent has attribute 'install_packages' with value 'this'
child has attribute 'install_packages+' with value 'that'
result is child having attribute 'install_packages' with value ['this', 'that']

This works with parent not having the attribute at all (the '+' is basically ignored), parent having value in attribute (value gets changed to list and new value is appended) and parent having value as list (new value is appended to the end)

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -3,11 +3,14 @@
 from __future__ import unicode_literals, absolute_import
 
 import pytest
-from fmf.utils import filter, FilterError
+from fmf.utils import filter, FilterError, FileError, TypeError
+from fmf.base import Tree
+
 
 # ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 #  Filter
 # ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
 
 def test_filter():
     """ Function filter() """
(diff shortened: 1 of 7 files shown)


jkrysl:
This new feature needs to be documented, where is the best place to put this? Elasticity paragraph in features.rst?

psss:
Thanks for the idea/implementation, Jakub. Looks quite useful. A few thoughts: The best place for the documentation is probably the section "Inheritance". It would be good to only allow appending attributes of the same type, for example:

* string ... concatenate
* list ... append
* number ... add

Auto-converting string from parent with string from child into list does not seem to be the best way. I would like to prevent any guessing here. Could you please extend your feature in this way?

Some technical suggestions: Could you squash your commits into a single one and follow [commit](http://fmf.readthedocs.io/en/latest/contribute.html#commits) recommendations? Thanks.

tests/test_utils.py
```diff
@@ -3,15 +3,19 @@
...
 
 # ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 #  Filter
 # ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+
 def test_filter():
     """ Function filter() """
     data = {"tag": ["Tier1", "TIPpass"], "category": ["Sanity"]}
+    print data
```
psss: A testing leftover to be removed?

jkrysl:
I thought your proposal through and I think what I am doing now i more consistent with the FMF design. If you have 2 values in *.fmf file, it gets stored as a list:
tag:
 - value1
 - value2
ends as tag=[value1, value2].
So I treated adding to tag the same way - converting it to list and adding new value. What you are proposing is inconsistent with this behaviour, as by adding the same value you will end with either string, list or int, when if you add it directly in metadata, you get always list:
tag: value1
/something
 tag+: value2
Should be the same as this:
/something
 tag:
  - value1
  - value2
Where in your proposal this would end up being type_of_value1(value1+value2)...and only sometimes. Do you still want me to change it?

Note: The way I coded it is much easier for me, as I simply expect either string (1 value) or list (2+ values), where with your proposal I would have to check the 1 value if it consists of more values and split it, than proceed with the same code as now.

jkrysl:
In most recent commit I added the ability to use tag+ with list using list.extend(), so now the behaviour should be consistent no matter the type of value (both child and parent), as it will always lead to a list of parent[:] + child[:].

psss:
Yes, I believe the text/number use case is valid and could be quite useful in the future. Let's take this example:

    time: 3
    description:
        This is an overall description.

    /child:
        time+: 3
        description+:
            Some test-specific details.

With your implementation we get this:

    test/child
    description: ["This is an overall description.", "Some test-specific details."]
    time: [3, 3]

Here's what I would intuitively expect from the "+" operator:

    test/child
    description: This is an overall description. Some test-specific details.
    time: 6

Strings should be probably joined by a new line. I'll look into this in more detail when I'm back from PTO.

jkrysl:
Let me give you an example of how it works now with the description:
```
# fmf --name vdo/kernel_modules/signature
vdo/kernel_modules/signature
description: There are tests for VDO

# fmf --name vdo/kernel_modules/signature
vdo/kernel_modules/signature
description: There are tests for VDO and Checking of signature of modules 'uds' and 'kvdo'.
```
In the second usage of fmf cli I added description+: "Checking of signature of modules 'uds' and 'kvdo'." As you can see the filtering naturally connects items in list to readable string.
Completely different story are ints and floats...

My usage of '+' is to nest setup blocks (and install requirements):
```
requires_setup: "setup_iscsi"
/nested_test:
  requires_setup+: "setup_vdo"
```
This way I run some tests on iscsi and than run some on vdo on the same iscsi. I do not have to setup iscsi again. It is much easier to have these setups in list than to check for some special characters.

Also there is the issue of identifying string in python2 and python3, because types:  py3(str) != py2(str); py3(str) = py2(basestring) = py2(str) + py2(unicode). To make the code compatible for both versions we would have to add a requirement of either 'six' or 'future' to be able to check for the same type.

I think we need to discuss this a bit more...

psss:
For the use case mentioned above the following could be used:

    requires_setup: [setup_iscsi]
    /nested_test:
      requires_setup+: [setup_vdo]

With the same number of characters ;-)

jkrysl:
This should be ready for merge.

PR #9, opened 2018-04-03
Add docs example for setting up storage

jkrysl:
Adding a new section to examples.rst with storage setup. It links to new [project on github](https://github.com/jkrysl/storage_setup), because it requires some extra tooling which should not be part of this project.

### DIFF
--- a/docs/examples.rst
+++ b/docs/examples.rst
@@ -1,4 +1,3 @@
-
 ======================
     Examples
 ======================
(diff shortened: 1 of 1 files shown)


psss:
Thanks for providing the example, Jakub. I did just some minor adjustments (e.g. adding a requirement to `setups/setup_*` was a bit confusing) and merged.

PR #20, opened 2018-04-20
fmf/cli.py: fix `os.path.dirname`  example invocation

clebergnu:
`dirname()` lives under `os.path` module, not under `os` directly.

Signed-off-by: Cleber Rosa <crosa@redhat.com>


psss:
Ah, thanks for catching this typo.

PR #24, opened 2018-04-25
WIP: Add example of real coreutils components from fedora

jscotka:
Example how to use metadata and standard test roles for coreutils package


psss:
Thanks for the example, Honzo. I think it's better to have this type of examples related to Fedora CI directly on the wiki. We already have something similar here:

https://fedoraproject.org/wiki/CI/Metadata/Examples

It shows different ways how FMF could be intergrated with the Standard Test Interface. But I think we will need to investigate the integration more. My current impression is that using the complex `--format` expressions directly in the `tests.yaml` files will not be the preferred way as it is quite confusing to the user.

Issue #25, opened 2018-05-15
Broken inheritance since v0.3


jkrysl:
Updating to FMF version 0.3 breaks the inheritance feature. All of the missing attributes in 0.3 are defined somewhere in the parents and they are not getting passed down to the child.

# 0.2:
$ fmf --name vdo/vdo_cli/various_commands/stop_start_rebuild_success
tests/vdo/vdo_cli/various_commands/stop_start_rebuild_success
component: vdo and kmod-kvdo
description: Testing 'vdo stop' and 'vdo start' with '--forceRebuild', commands should succeed
requires_install: vdo and kmod-kvdo
requires_setup: vdo/setup/storage/setup_* and vdo/setup/create_vdo
test: stop_start_rebuild.py
tester: me...
tier: 1
time: 1 min
vdo_name: vdo_test

# 0.3:
$ fmf --name vdo/vdo_cli/various_commands/stop_start_rebuild_success
tests/vdo/vdo_cli/various_commands/stop_start_rebuild_success
description: "Testing 'vdo stop' and 'vdo start' with '--forceRebuild', commands should succeed"
test: 'stop_start_rebuild.py'
tier: 1

jkrysl:
This seems to be related to 70b6207f71b4e00622f55e1420c3e57e061a71a0, as I am defining the same leaf (quite deep in hierarchy) from 2 places:
```
# fmf --name vdo/vdo_cli/various_commands/stop_start_rebuild_success
tests/vdo/vdo_cli/various_commands/stop_start_rebuild_success
description: "Testing 'vdo stop' and 'vdo start' with '--forceRebuild', commands should succeed"
something: True
test: 'stop_start_rebuild.py'
tier: 1

# mv vdo.fmf ..

# fmf --name vdo/vdo_cli/various_commands/stop_start_rebuild_success
tests/vdo/vdo_cli/various_commands/stop_start_rebuild_success
component: ['vdo', 'kmod-kvdo']
description: "Testing 'vdo stop' and 'vdo start' with '--forceRebuild', commands should succeed"
requires_install: ['vdo', 'kmod-kvdo']
requires_setup: ['vdo/setup/storage/setup_*', 'vdo/setup/create_vdo']
test: 'stop_start_rebuild.py'
tester: 'me...'
tier: 1
time: '1 min'
vdo_name: 'vdo_test'

# cat ../vdo.fmf
/vdo_cli/various_commands/stop_start_rebuild_success:
  something: True

[...]

PR #28, opened 2018-05-16
[WIP] Ensure inheritance of all parent data no matter where

jkrysl:
Fixes issue #25 

I did not find a way to fix this without the duplication. It does not work using only one of them (or moving the 2nd from MAIN file loading). The one in MAIN file loading is for inheritance through the tree while the one in update() is for direct inheritance from the parent.

### DIFF
--- a/examples/inheritance/one.fmf
+++ b/examples/inheritance/one.fmf
@@ -1,0 +1,5 @@
+one_root: 1
+/two:
+  two_root: 1
+/two/three:
+  three_root: 1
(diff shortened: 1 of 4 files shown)


jkrysl:
I added a test to hit the issue. It appears only when there is file with name other than `main.fmf` that contains data of the same leaf as somewhere deeper in the tree. In that case it completely ignores any data in between.
The patch so far fixes it by 1 level from each side, but the middle is still broken. There has to be better was to fix this...

psss:
This should be fixed by 6d3da7f.

PR #33, opened 2018-05-29
initial version of testsets filtering and formatting via configs

jscotka:
add possibility to store filters and formatting in configs.
@psss what do you thing about that.
and probably If we'll agree on this feature, I suggest move part of code from `cli` to `base` to `prune` function or create separated one for this purpose. It'll fix semantics of what **test sets** means

### Usecase
 * I would like to be able to store configuration for  tier tests (each component could have `tier1` testset defined by own and then you will rely that this filter is there)
 * I would like to simplify usage of FMF for some std purposes like have filters for fast and slow test, then use it in makefiles
 * initial reason was, that I had to implement similar feature in MTF tool: https://github.com/jscotka/meta-test-family/blob/03b4a50442e735efd9db215caa102674604d104d/mtf.conf.d/mtf.yaml#L104
 * rulesets in COLIN project are very similar concept: https://github.com/user-cont/colin/blob/master/rulesets/fedora.json 

### Examples:
```
/testset1:
  keys:
    - test
  names: []
  filters:
    - "tags:Tier1"
  format: "{}/{}/{}\n"
  values:
    - "root"
    - "data.get('path',name)"
    - "data['test']"
```

```
$ fmf examples/conf_formatting/ --testset testset1 
/home/jscotka/git/fmf/examples/conf_formatting/wget/download/test/runtest.sh
/home/jscotka/git/fmf/examples/conf_formatting/wget/protocols/ftp/runtest.sh
/home/jscotka/git/fmf/examples/conf_formatting/wget/protocols/http/runtest.sh
/home/jscotka/git/fmf/examples/conf_formatting/wget/protocols/https/runtest.sh

$ fmf examples/conf_formatting/ --testset testset1 --name download
/home/jscotka/git/fmf/examples/conf_formatting/wget/download/test/runtest.sh

$ fmf examples/conf_formatting/ --testset testset1 --format "{}\n" --value name
conf_formatting/wget/download/test
conf_formatting/wget/protocols/ftp
conf_formatting/wget/protocols/http
conf_formatting/wget/protocols/https
```


psss:
I think I understand the use case but I'm not completely sure whether this feature fits the scope of FMF. The command line tool was intended for quick inspection and experimenting with the metadata. My impression is that something like this would be much easier done directly through the Python API. But I'm open for discussion. If we decide to add this cli feature I would suggest to identify the test set using node names / identifiers (use `find()` instead of prunning the tree). Also combining multiple test sets seems to be quite specific use case.

jscotka:
@psss Yep, I agree. we can move it to `filter` function. It sounds like best idea, and I also agree, that supprort in cli is not so imrpotant, just good to have this feature inside FMF, to show this abilities and avoid to so similar implementation in own tools.

I was not sure if we would like to support store there filtering or also presenting of data. I've mixed it together as example how it can be simple and very elegant how to use it, but I agree that probably just filtering is enough to store there.

psss:
As explained in [comment above](https://github.com/psss/fmf/pull/33#issuecomment-396260494) this seems to be out of scope for the simple command line inspection.

Issue #34, opened 2018-06-04
node referencing and inheritance


jscotka:
for one project I would like to see some semantics to be able to reference nodes via name (this simple reference should be enough)
see example code:
```
/maintainer_label:
  test: script.py
  tags: ["a", "b"]

/sanity
  /from_tag_not_latest
    test: anotherscript.py

/tests
  /checks:
    /check1@maintainer_label:
      tags+: ["required"]
    /check2@/sanity/from_tag_not_latest:
      tags+: ["required"]
      usable_targets: ["dockerfile"]
```
what will use special symbol `@` to use this item name as `sources` and  data will be copied from this item instead of parent in tree structure.
so that list item `/tests/checks/check1` will contain:
```
  test: script.py
  tags: ["a", "b", "required"]
```

does that make sense?

Possible issues
 * wait to initialize all  elements without @ items  and then these
 * allow add there non trivial references, like `a@b@c ` and allow also references to other refs items
 *  could leads  to circular deps -> should be able to detect cycles
 * will be data copied, or "really" references, so that change of one will change also second one

jscotka:
my case solved by: https://github.com/psss/fmf/pull/36
I've decided to solve it as postprocessing step, because it could be easier to find that base objects are constructed already

jkrysl:
I am not sure this is the best idea, from point of FMF this seems quite bad. You are basically breaking the tree structure, where some leaves belong to a branch but do not take anything from it, instead they are virtually moved to different branch. 
So what prevents you from moving those tests where they actually belong and keep the inheritance sane and intact?

jscotka:
@jkrysl Tree structure is not broken, just one thing what is chaged is that these `data` elements are inherited from reference, not from parent items and also copied sources, so that easy to debug this.
But I understand you PoV.

My usecase is, that someone, wants to create virtual hierarchy based on some existing one.
(maybe could be used for linking (merging) test coverage  with real cases)

AloisMahdal:
+1 to @jkrysl; this seems risky at best.

If you have some project-specific mapping, that's fine (most if not all projects will have some own rules), but you should always do that on project-specific tooling.  Maybe I'm not understanding what you mean, but it seems like something you should write your own tools for that; FMF should really be just about discovering and collecting plain data -- not giving meaning or semantics.

AloisMahdal:
And by the way, what if I want to have key with `@` in it?

jscotka:
@AloisMahdal 
[...]

PR #36, opened 2018-06-05
reference resolver via @ in names

jscotka:
#### main.fmf
```
/downstream-checks:
  /check1@wget.recursion.fast:
    test+: " --param 1"
  /check2@deep:
     tags+: ["fast", "required"]
```
#### result
for command: `$ fmf examples/referencing/ --name downstream `
```
referencing/downstream-checks/check1@wget.recursion.fast
depth: 10
description: Check recursive download options
path: wget/recursion
tags: Tier2
test: runtest.sh --param 1
tester: Petr Šplíchal <psplicha@redhat.com>
time: 1 min

referencing/downstream-checks/check2@deep
depth: 1000
description: Check recursive download options
path: wget/recursion
tags: Tier2, fast and required
test: runtest.sh
tester: Petr Šplíchal <psplicha@redhat.com>
time: 10 min
```

### Test location issue
This leads to another complexity where is `test` :-)
But it is not so big problem because you know, you uses it, and for example I'll use `sources[-2]` in case `name` contains `@`


psss:
As explained in #34 it is currently out of scope to implement such a complex referencing. Closing for now. We can possibly revisit this later.

PR #41, opened 2018-06-18
improvements for external tools.

jscotka:
few improvements what helps me much.


jscotka:
point of this moving this out is, that I want to implement these virtual case structure and refecences via `@` like:  https://github.com/psss/fmf/pull/36
so that I need to have possibility to use this internal semantics merging, a do not implement this in my tool once more time.
I want to use same `+` symbol in same way.

PR #48, opened 2018-10-01
Ignore directories with no metadata defined

psss:
There might be additional directories included in the directory structure which are not related to test metadata. We should not create tree objects for these.

Here's one real-life example from the libsemanage tests:
https://src.fedoraproject.org/tests/selinux/blob/master/f/libsemanage/sanity-tests

Similarly it (probably) does not make sense to store empty leaves defined by virtual structure (attributes starting with slash). Attached patch should resolve both.


psss:
@thrix, @jscotka, @AloisMahdal, @jkrysl, what are your thoughts on this?

AloisMahdal:
Seems like a perfectly logical behavior.

thrix:
yep, I like it, so :+1:

PR #58, opened 2019-08-10
Enable packit

psss:


TomasTomecek:
Trying this now locally: I hope I'd be able to get back to you tomorrow morning since I have to leave now.

I also ran into this one using packit 0.6.0: https://github.com/packit-service/packit/issues/524

TomasTomecek:
```diff
diff --git a/.packit.yaml b/.packit.yaml
index 8ca85d0..a5a876f 100644
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -15,6 +15,4 @@ jobs:
       - fedora-29-x86_64
 
 actions:
-  create-archive:
-    - make tarball
+  create-archive: make tarball
diff --git a/Makefile b/Makefile
index 1ef44bd..31204d1 100644
--- a/Makefile
+++ b/Makefile
@@ -43,7 +43,8 @@ source: clean
        mkdir -p $(TMP)/$(PACKAGE)
        cp -a $(FILES) $(TMP)/$(PACKAGE)
 tarball: source man
-   cd $(TMP) && tar cfz SOURCES/$(PACKAGE).tar.gz $(PACKAGE)
+ cd $(TMP) && tar cfz $(PACKAGE).tar.gz $(PACKAGE)
+ mv $(TMP)/$(PACKAGE).tar.gz ../
 rpm: tarball
        rpmbuild --define '_topdir $(TMP)' -bb fmf.spec
 srpm: tarball
```

It should work with this patch.

Btw I stumbled upon another bug in packit 0.6.0 https://github.com/packit-service/packit/pull/526

psss:
This will not work for me as it breaks other targets (make srpm, make rpm). It's not possible to provide multiple commands to the `create_archive` option? Or you just need to have the tarball stored one level higher in the directory path (as the patch suggests)?

TomasTomecek:
> This will not work for me as it breaks other targets (make srpm, make rpm). It's not possible to provide multiple commands to the `create_archive` option?

so can you wrap the two commands in a new make target?

> Or you just need to have the tarball stored one level higher in the directory path (as the patch suggests)?

Sadly there is bug in create-archive that it can't accept the list of commands now: packit-service/packit#524 The problem is that we read the path to the archive from stdout, so if it's a list, we don't know where to read it from.

TomasTomecek:
Darn it, I'll take a look on Monday.

psss:
@TomasTomecek is there any issue with the spec?

TomasTomecek:
It is, actually :D you are not using `Source0` as the primary source. I'll create a packit issue to track this, but I've never seen a spec file which would reference primary archive as Source and not as Source0. But I haven't seen that many spec files.

https://github.com/psss/fmf/pull/59 With these changes I was able to rebuild fmf locally.

TomasTomecek:
Party time!

psss:
Hooray! Just one last thing: Moving the tarball breaks "make srpm". Could we modify `print-tarball-path` instead to provide full path?

TomasTomecek:
so replace `make srpm` with `packit srpm` ;)

I don't know to be honest: I've never tried to put absolute paths inside a spec file.

psss:
@TomasTomecek, thanks much for your help!

TomasTomecek:
my pleasure

PR #68, opened 2020-01-09
Get metadata for unique id [fix #67]

### LINKED ISSUE #67: Get metadata for unique id (remote and cache support)
Add method to get metadata based on #52 so tools as psss/tmt discover do not need to figure out git clone or use cache & run fmf on their own.

E.g. Checkout remote under `XDG_CACHE_HOME` and keep it there for further use.

lukaszachy:


psss:
Thanks for the patch, @lukaszachy. Looks good. I'm just thinking about the best / the most consistent way how to integrate this new functionality into existing code. Additional function seems fine but as its result is supposed to be a `Tree` instance this could be also handled directly in the constructor.

But there is the `data` parameter... What about updating the `__init__()` parameters in this way?

    def __init__(self, identifier=None, data=None, name=None, parent=None):

Data is rather used internally for creating children, we would keep it, but as the first parameter there would be `identifier` which could be both `string` as path (backward-compatible) or `dict` as the full fmf identifier. Trees would be initialized in this way:

    tree = Tree('/local/path/to/metadata/tree')
    tree = Tree(dict(url='https://git.repo'))
    tree = Tree(dict(url='https://git.repo', path='/path/in/the/repo'))
    tree = Tree(dict(url='https://git.repo', path='/path/in/the/repo', revision='devel'))

This could (possibly in the future) even support picking the specific node by providing the node name as well. I think this approach could come in handy.

    tree = Tree(dict(url='https://git.repo', path='/path/in/the/repo', revision='devel', name='/node/name'))

What do you think, @lukaszachy? @thrix, @jkrysl, @jscotka any comments or concerns?

lukaszachy:
> But there is the `data` parameter... What about updating the `__init__()` parameters in this way?
> 
> ```
> def __init__(self, identifier=None, data=None, name=None, parent=None):
> ```

Current implementation is
```
def __init__(self, data, name=None, parent=None):
```

So if one doesn't use named parameter `data` (accepting string or dict) in current code the backward compatibility break. I am not sure how many people are using it now in such way though.

While it is a very nice one-stop method, wouldn't be there too much functionality hidden in the constructor? 

On the other hand, if one can specify node name in the Tree constructor: How about allowing names for local directories as well?

psss:
> So if one doesn't use named parameter `data` (accepting string or dict) in current code the backward compatibility break. I am not sure how many people are using it now in such way though.

Yes, this is slightly backward-incompatible but I believe feeding data directly would be a very rare case if any.

> While it is a very nice one-stop method, wouldn't be there too much functionality hidden in the constructor?

I would still keep the `checkout_remote()` function in utils. Otherwise I don't think constructor would handle many different scenarios.

> On the other hand, if one can specify node name in the Tree constructor: How about allowing names for local directories as well?

Hm, this is probably misleading (if I understand you correctly). The `name` parameter is used for giving a name to a child tree node not for selecting nodes. Or did you mean something else?

lukaszachy:
> I would still keep the `checkout_remote()` function in utils. Otherwise I don't think constructor would handle many different scenarios.

OK
> 
> > On the other hand, if one can specify node name in the Tree constructor: How about allowing names for local directories as well?
> 
> Hm, this is probably misleading (if I understand you correctly). The `name` parameter is used for giving a name to a child tree node not for selecting nodes. Or did you mean something else?

I meant `name` from `remote object reference`. So we can limit single test node in similar way for both remote and local fmf storage:
> This could (possibly in the future) even support picking the specific node by providing the node name as well. I think this approach could come in handy.

Now you need to use:
```
tree_data = Tree('/path/in/the/repo')
test_itself = tree_data.prune(names=['/test/name'])[0]
```
So if I could use `remote reference` to get me single test by calling `Tree()` why not to allow me to get single test by calling `Tree` also for locals? 
E.g. (option names need tweaking a bit)
```
# For local
test_itself = Tree('path/in/the/repo', limit_to_name='/test/name'])
# For remote
test_itself = Tree(dict(url='github', path='path/in/the/repo', name='/test/name']))
```

psss:
Thanks for clarification. Yes, completely agreed. Something like this could be a relatively common use case. What about just using the remote identifier without `url` key?

    test = Tree({path: '/local/path', name: '/test/name'})

Seems better than introducing a new parameter. Oterwise definitely plus one for this.

psss:
Seems we have an agreement on the implemenation. @lukaszachy are you going to update the pull request or shall I look into that?

lukaszachy:
> Seems we have an agreement on the implemenation. @lukaszachy are you going to update the pull request or shall I look into that?

I'll update the pull request soonish.

lukaszachy:
Close & reopen to trigger travis again as it didn't recognized [pass](https://travis-ci.org/psss/fmf/builds/640104302)

lukaszachy:
CI passing - feedback & review appreciated, @psss when you have time

fmf/utils.py
```diff
@@ -500,6 +501,57 @@ def enabled(self):
...
+                See base.Tree.node for description
+                Key 'name' is ignored.
+    """
+
+    default_for_rev = 'master'
+
+    destination = os.path.join(
+        os.path.join(os.path.expanduser(os.environ.get('XDG_CACHE_HOME', '~/.cache')), 'fmf'),
+        remote_ref['url'].replace('/', '_'),
+        remote_ref.get('ref', default_for_rev).replace('/', '_'),
```
psss: Is it necessary to create a separate directory for each `ref`? Would it be ok to have just a single directory for each `url` under the cache and just use `git checkout` for switching refs? Especially for larger repositories this could save some space.

psss: Or is this needed to prevent possible race conditions between multiple processes?

lukaszachy: I was trying to avoid git checkout every time. But doing it every time might be better

psss: I see. Ok, checkout always seems to me a bit better to save resources with fetching repos.

lukaszachy:
I need to check this PR adheres to definition from #71

lukaszachy:
Implementation should follow the docs for ``fmf ￼identifier``.
I didn't manage to implement locking as it is new stuff for me and I need to understand options first.

psss:
Thanks much for looking into this. Merged with some adjustments into master.

PR #71, opened 2020-02-06
Define the full fmf identifier [fix #52]

### LINKED ISSUE #52: Define how to reference remote objects
Currently each node in the metadata tree structure has a unique identifier called `name` so it is possible to reference individual objects from other attributes. This can be useful for example for mapping test coverage where tests are referenced from requirements.

This works nicely within a single metadata tree. However, we need to define a clear way how to reference objects which are outside of the current metadata tree. For this we could combine repository url with object name using a defined separator. 

Also, in order to prevent too long identifiers and repeating urls it would be good to provide a way how to define a common prefix or remote repository shortcut. Perhaps it could work in a similar way how remotes like origin are defined in git.

psss:
Based on the discussion in #52 proposing the definition of the full fmf identifier. For now I've omitted the `commit` key. If it really will be necessary, I believe we can add it later. @thrix, @lukaszachy, @pvalena, @t184256, @comps, @kkaarreell, @hegerj, please have a look. Thanks.

### DIFF
--- a/docs/concept.rst
+++ b/docs/concept.rst
@@ -116,3 +116,64 @@ directory. Below you can see node names for different files:
     +-------------------------------+-----------------------+
     | wget/download/smoke.fmf       | /download/smoke       |
     +-------------------------------+-----------------------+
+
+
+Identifiers
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Node names are unique across the metadata tree and thus can be
+used as identifiers for local referencing across the same tree. In
+order to reference remote fmf nodes from other trees a full ``fmf
+identifier`` is defined as a dictionary containing keys with the
+following meaning:
+
+url
+    Git repository containing the metadata tree. Use any format
+    acceptable by the ``git clone`` command. Optional, if no
+    repository url is provided, local files will be used.
+ref
+    Branch, tag or commit specifying the desired git revision.
+    This is used to perform a ``git checkout`` in the repository.
+    Optional, by default the ``master`` branch is used.
+path
+    Path to the metadata tree root. Should be relative to the git
+    repository root if ``url`` provided, absolute local filesystem
+    path otherwise. Optional, by default ``.`` is used.
+name
+    Node name as defined by the hierarchy in the metadata tree.
+    Optional, by default the parent node ``/`` is used, which
+    represents the whole metadata tree.
+
+Here's a full fmf identifier example::
+
+    url: https://github.com/psss/fmf
+    ref: 0.10
+    path: /examples/wget
+    name: /download/test
+
+Use default values for ``ref`` and ``path`` to reference the
+latest version of the smoke plan from the ``master`` branch::
+
+    url: https://github.com/psss/fmf
+    name: /plans/smoke
+
+If desired, it is also possible to write the identifier on a
+single line as supported by the ``yaml`` format::
+
+    {url: "https://github.com/psss/fmf", name: "/plans/smoke"}
+
+Let's freeze the stable test version by using a specific commit::
+
+    url: https://github.com/psss/fmf
+    ref: f24ef3f
+    name: /tests/basic/filter
+
+Reference a smoke plan from another metadata tree stored on the
+local filesystem::
+
+    path: /home/psss/git/tmt
+    name: /plans/smoke
+
+Local reference across the same metadata tree is also supported::
+
+    name: /plans/smoke


docs/concept.rst
```diff
@@ -116,3 +116,50 @@ directory. Below you can see node names for different files:
...
+Let's freeze the stable test version by using a specific commit::
+
+    url: https://github.com/psss/fmf
+    ref: f24ef3f
+    name: /tests/basic/filter
+
+Reference another repository on the local filesystem::
+
+    url: file:///home/psss/git/tmt
+    name: /plans/smoke
```
t184256: Does that one refer to `master` or to the working tree?

thrix:
@happz also I think you should look at this ...

docs/concept.rst
psss: Unless specified, `ref` defaults to master. So this would be `master`. Would you expect something else?

thrix: good point ... and I do not have a good answer what it should be .... :(

t184256: Me being myself, I may be puzzled why my uncommitted working tree changes are not picked up. I don't see a nice solution, but it should be probably explicitly stated that it will clone and checkout master, not use the working tree.

psss: We could also make this example a little bit more clear and specific by adding `ref: HEAD` and stating it would use content from `HEAD`. Not sure how something like this would be common though.

comps: FTR; A working tree is never a part of a repository. No git commands can "fetch" (or get in any way) a remote working tree. One repository can also be referenced from multiple working trees, each with their own index (stage).
It's only `gitk` / `tig` that propagates this myth via the "Uncommited changes", making them seem like a commit.

@psss For maximum correctness, the remote `ref` should default to `HEAD`, not `master`, as a remote `HEAD` points to what people (and github) usually call a "default branch".

psss: Local filesystem example desription updated to be explicit about the branches. @t184256, does it now look ok to you?

thrix: for me it looks good

t184256: @psss, IDK, now @comps made me wonder whether implicit HEAD would fare better than master or introduce even more confusion. But my original complaint is definitely resolved now, as it's clear it will clone from it and not use the working tree.

psss:
Updated based on the irc feedback. Please, re-review. Thanks.

thrix:
@psss I will remove the f29 chroot, it is gone now .... we should switch to fedora-*

thrix:
> @psss I will remove the f29 chroot, it is gone now .... we should switch to fedora-*

maybe in another PR

thrix:
Submitted to #72

pvalena:
@lukaszachy,  git actually fully supports local URI:

https://mirrors.edge.kernel.org/pub/software/scm/git/docs/git-clone.html#URLS

so `file://path/to/file` is a valid value for `git clone`, as long as you're on the correct filesystem.

psss:
> git actually fully supports local URI

Yes, but the thing is we would like to support working with uncommitted changes as well. With the latest version of the spec we can do both:

    url: file:///local/repo
    name: /plans/smoke

which will fetch the committed changes and use the `master` branch but also this:

    path: /local/repo
    name: /plans/smoke

which will use anything what is currently in the working directory, including uncommitted changes.

Thanks much to all for your review and valuable feedback!

comps:
I don't know, that seems to me like a lot of magic in one place.
1. Using two completely different retrieval mechanisms based on path prefix.
1. Using `master` as default (hint: `git clone` doesn't clone master by default).
1. Using something that isn't a `ref` (commit hash) inside `ref`.

Sounds to me like you would benefit from a more general system that could reference remote metadata by (explicitly) choosing between protocols, ie. "local directory" (would work on uncommitted changes too), "local archive" (could reference files inside a `zip`), "git repository", etc.

psss:
> 1. Using two completely different retrieval mechanisms based on path prefix.

The difference is not only the prefix: You either specify `url` (first example above) then git repository is fetched, or you don't and only use `path` (second example above) in which case no git action are performed. Seems quite clear and distinct to me. 

> 2. Using `master` as default (hint: `git clone` doesn't clone master by default - try cloning https://github.com/comps/gitshop). Maybe rephrase to "if unspecified, default branch is used".

The intention is really to use `master` branch by default, not the default branch. Not sure how many people know what exactly the default branch means, while `master` is completely clear.

One more thought: We want to re-use cloned/cached repositories (imagine scheduling 100 tests from the same repo). So `git clone` would be done only once and the only switching `ref` if provided. Is it possible to switch somehow to the default remote branch once `HEAD` is updated after `git checkout`?

> 3. Using something that isn't a `ref` (commit hash) inside `ref`.

I see. Not sure here. For now it really seems to me that one more additional key is not necessary as the value will be used with `git checkout` anyway.

Issue #78, opened 2020-05-20
Cache update not working properly


Athwale:
HEAD is not advanced after git fetch in the fmf repo cache after tests are modified in test repo.
The update should happen every 20 minutes, but does not.

lukaszachy:
@psss could you take a look, please?

PR #79, opened 2020-05-20
Checkout origin/<ref> instead of <ref>

lukaszachy:
Attempt to solve #78


lukaszachy:
Hopefully last iteration, thank you @psss for review via IRC.

PR #88, opened 2020-11-18
Remove the duplicate build job from packit config

psss:


psss:
@lachmanfrantisek, no update from packit if the extra section is removed.

lachmanfrantisek:
> @lachmanfrantisek, no update from packit if the extra section is removed.

But it works well without that change, isn't it?

psss:
@lachmanfrantisek, yes, but then you get two results for each distro which might be confusing.

lachmanfrantisek:
@psss Yes, I just wanted to make sure that it (somehow) works until we fix it.

psss:
I see. Thanks for looking into this, @lachmanfrantisek.

PR #90, opened 2020-11-23
Raise CannotDecide for missing version part

### LINKED ISSUE #89: Matching distro against a context with distro name only
If there is no distro version specified in the context, comparing against distro specific version gives `False`:

    >>> import fmf
    >>> fmf.context.Context(distro='fedora').matches('distro < fedora-33')
    False

@lukaszachy, shouldn't this raise a `CannotDecide` exception instead?

lukaszachy:
If you want to compare you need to have version defined on both sides.
Alphabetical order doesn't make sense for these use cases

Resolves #89

Notable difference is to previous behavior is that `foo > foo` is `CannotDecide`. Otherwise not such big difference, see changed tests.

### DIFF
--- a/docs/context.rst
+++ b/docs/context.rst
@@ -69,11 +69,7 @@ adjusted. For example it can be arch, distro, component, product
 or pipeline in which we run tests and so on.
 
 Each value is treated as if it was a component with version. Name
-of the dimension doesn't matter, all are treated equally. For some
-dimensions, such as ``arch``,  only comparison for equality makes
-sense. Note that the implementation does not raise an error when
-comparing ``aarch64 > x86_64``. In this case the alphabetical
-order is defined, you were warned.
+of the dimension doesn't matter, all are treated equally.
 
 The characters ``:`` or ``.`` or ``-`` are used as version
 separators and are handled in the same way. The following examples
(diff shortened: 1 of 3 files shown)


lukaszachy:
Quite large change in the end. See docs and tests.

psss:
The original example which triggered #89 now gives a `True`:

    >>> fmf.context.Context(distro='fedora').matches('distro < fedora-33')
    True

Using major version comparison ends up with a `CannotDecide`:

    >>> fmf.context.Context(distro='fedora').matches('distro ~< fedora-33')
      File "<stdin>", line 1, in <module>
      File "/home/psss/git/fmf/fmf/context.py", line 500, in matches
        raise CannotDecide()  # It's up to callee how to treat this
    fmf.context.CannotDecide

Could you please remind me why the first case should not result into a `CannotDecide` exception as well?

lukaszachy:
> Could you please remind me why the first case should not result into a `CannotDecide` exception as well?
Because missing version parts of left side are treated  as if they were lower than right side (https://github.com/psss/fmf/pull/90/files#diff-c5fcdc91598d4110dcf7dd81ddf7bdcde23ac47086864e281a7538d7ad879901R100)

Motivation is that if some package doesn't define all version pars. e.g. foo-2.0 vs foo-1.2.3.hotfix.
If you want to be sure to compare with 'major' version defined you should use ~.

psss:
Ah, I see. So that makes a perfect sense if the context has at least some version defined. But if is there is no version?

    >>> fmf.context.Context(package='foo').matches('package > foo-1')
    False
    >>> fmf.context.Context(package='foo').matches('package < foo-1')
    True

I find this behaviour rather confusing/unpredictable. Shall we perhaps require at least some version on the left side if version is provided on the right side?

lukaszachy:
> Ah, I see. So that makes a perfect sense if the context has at least some version defined. But if is there is no version?
> 
> ```
> >>> fmf.context.Context(package='foo').matches('package > foo-1')
> False
> >>> fmf.context.Context(package='foo').matches('package < foo-1')
> True
> ```
> 
> I find this behaviour rather confusing/unpredictable. Shall we perhaps require at least some version on the left side if version is provided on the right side?

Easy to do. I'll fix.

psss:
Thanks a lot! Rebased on the latest master and adjusted a bit the docs plus a bit of style cleanup.

psss (approved):

PR #91, opened 2020-11-25
Adjust 'make test' to work under virtualenv

psss:
Explicitly provide pytest config to suppress warnings.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,13 +17,11 @@ tmp:
 
 # Run the test suite, optionally with coverage
 test: tmp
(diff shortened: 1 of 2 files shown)


lukaszachy (approved):
LGTM, it just tests from installed versions so `pip install -e .` in virtualenv will be necessary for development. BUt that is a best practice, right?

psss:
> virtualenv will be necessary for development. BUt that is a best practice, right?

Yes, exactly. Thanks for spotting the missing config!

PR #92, opened 2020-11-25
Update the adjust continue key default to True

psss:
By default all rules are evaluated.
Docs updated and moved under features.


lukaszachy (approved):
LGTM

PR #93, opened 2021-01-05
Implement recursive merging of data

jscotka:


psss:
Thanks for implementing this. Looks good. However, if we extend the merge functionality for the `+` operator we should do the same for the `-` suffix as well to make the behaviour consistent.

jscotka:
> Thanks for implementing this. Looks good. However, if we extend the merge functionality for the `+` operator we should do the same for the `-` suffix as well to make the behaviour consistent.

Hi,
I don't think so, or better to say, `-` has completely another implementation than `+`, so from my perspective it was never consistent from perspective of reverse operation.
as I've looked into implementation of recursivity for dictionaries inside `-`. For `-` it is supported just list of arguments to `pop` from `dict`. But I can try to implement go deep into dict structure if it is there if `-` is used.

jscotka:
@psss just maybe add testcase like, just to prove, that it is able to mix - + leaves together inside one dict:  
```
    very+:
        deep+:
            dict+:
                y-: 1
                x+: 1
                z: 1
         another-: 1
```

psss:
> maybe add testcase like, just to prove, that it is able to mix - + leaves together inside one dict

Good point, makes sense. Adjusted.

jscotka:
> > maybe add testcase like, just to prove, that it is able to mix - + leaves together inside one dict
> 
> Good point, makes sense. Adjusted.

thanks a lot. it will help to avoid any possible regression when rewriting internal parts of this logic. if any.

psss:
Good, thanks, merged.

PR #98, opened 2021-01-19
Implement a new method copy() for cloning nodes

psss:
This can useful when adjusting metadata multiple times. Instead of
parsing files and growing trees multiple times it is now possible
to easily create a deep copy of the node and its subtree.

### DIFF
--- a/fmf/base.py
+++ b/fmf/base.py
@@ -572,3 +572,19 @@ class Tree(object):
                 "No tree node found for '{0}' reference".format(reference))
         # FIXME Should be able to remove .cache if required
         return found_node
+
+
+    def copy(self):
+        """
+        Create and return a deep copy of the node and its subtree
+
+        It is possible to call copy() on any node in the tree, not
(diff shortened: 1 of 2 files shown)


psss:
@lukaszachy, @t184256, here's the implementation of the new `Tree.copy()` method which allows cloning subtrees without duplicating the whole tree. Please, have a look. Thanks.

t184256:
Yeah, that's what I wanted. Maybe I'm paranoid, but I'd also suggest:

1) a test that changes in child nodes not propagating either way, to ensure the copy is deep enough
2) a test that adjusts a subtree and checks that changes didn't spread

psss:
> * a test that changes in child nodes not propagating either way, to ensure the copy is deep enough

I've added a new test `test_node_copy_child` which should cover this.

> * a test that adjusts a subtree and checks that changes didn't spread

I think that should be covered by `test_node_copy_subtree`. Or, perhaps, you meant something else?

lukaszachy (approved):

Issue #101, opened 2021-01-25
Support for storing updated metadata to disk


psss:
Thanks to virtual hierarchy and elasticity it is not an easy task to store an updated metadata back to disk because the data can be stored at multiple locations. Support for merging attributes makes the task even harder. It would be nice if `fmf` supported this functionality and stored updated attributes at to their appropriate locations in a comfortable and automated way.

psss:
Pasting several @jscotka's implementation-related remarks from psss/tmt/issues/427:

It reminds me, that similar issue could be with yaml loader inside FMF, and my perspective on this issue is that most of work would be done on FMF side and TMT should be able to use FMF "saving" abilities then instead of own way.

FMF has data which node is stored in which files and probably for this case you want to update as latst file as possible. little trouble is that FMF does not remember raw data, just merged data +- and directory structure

- so probably it need to add to FMF some attributes, which exact part of dict inside file is used for this node (now there are just list of files)
- then just updating this node inside linked file, by new attribute (have to load this pure file again, to not dump merged `+` `-` items)
- then just dump the updated data to files if there are changes (maybe some attribute for changes?)

This way could be harder to implement (but not so much, just extending fmf node attributes and then just use it and dump), but cleaner for future usage and implementing more complicate updating/optimizing of fmf files.

PR #102, opened 2021-01-29
Add support for storing modified metadata to disk

jscotka:
it implements modification of fmf files:
solves: https://github.com/psss/fmf/issues/101

Probably it causes reordering of output file, but it is another issue and solution exists, but it little bit tricky, based on improving yaml parser for using OrderedDict: https://stackoverflow.com/questions/13297744/pyyaml-control-ordering-of-items-called-by-yaml-load with some solution like: https://github.com/wimglenn/oyaml

EDIT:
problem with reordering solved by using `sort_keys=False` for `yaml.safe_dump`

### DIFF
--- a/fmf/base.py
+++ b/fmf/base.py
@@ -610,76 +610,80 @@ class Tree(object):
         self.parent = duplicate.parent = original_parent
         return duplicate
 
-    def _get_raw_data_node(self):
+    def _locate_raw_data(self):
         """
-        Get tuple
-          - node where _raw_data are stored
-          - list if keys for current node inside raw data
+        Detect location of raw data from which the node has been created
+
+        Find the closest parent node which has raw data defined. In the
+        raw data identify the dictionary corresponding to the current
+        node, create if needed. Detect the raw data source filename.
+
+        Return tuple with the following three items:
+
+        node_data ... dictionary containing raw data for the current node
+        full_data ... full raw data from the closest parent node
+        source ... file system path where the full raw data are stored
+
         """
-        raw_data_node = self
-        id_list = list()
+        # List of node names in the virtual hierarchy
+        hierarchy = list()
+
+        # Find the closest parent with raw data defined
+        node = self
         while True:
-            raw_data = raw_data_node._raw_data
-            if raw_data:
-                return raw_data_node, id_list
-            elif not raw_data_node.parent:
-                raise utils.RootError("Cannot find root node with _raw_data, very strange")
-            else:
-                id_list.insert(0, "/" + raw_data_node.name.rsplit("/")[-1])
-                raw_data_node = raw_data_node.parent
+            # Raw data found
+            full_data = node._raw_data
+            if full_data:
+                break
+            # No raw data, perhaps a Tree initialized from a dict?
+            if not node.parent:
+                raise utils.GeneralError(
+                    "No raw data found, does the Tree grow on a filesystem?")
+            # Extend virtual hierarchy with the current node name, go up
+            hierarchy.insert(0, "/" + node.name.rsplit("/")[-1])
+            node = node.parent
 
-    def _get_node_raw_data(self, default=None):
-        """
-        Get dictionary of _raw_data for current node
-        in case of None add to the current node key default element
-        """
-        raw_data_node, id_list = self._get_raw_data_node()
-        raw_data = raw_data_node._raw_data
-        for key in id_list:
-            if default is not None and raw_data[key] is None:
-                raw_data[key] = default
-            raw_data = raw_data[key]
-        return raw_data
+        # Localize node data dictionary in the virtual hierarchy
+        node_data = full_data
+        for key in hierarchy:
+            # Create a virtual hierarchy level if missing
+            if key not in node_data:
+                node_data[key] = dict()
+            # Initialize as an empty dict if leaf node is empty
+            if node_data[key] is None:
+                node_data[key] = dict()
+            node_data = node_data[key]
 
-    def modify(self, *data, method="update"):
-        """
-        Modify node metadata (original data stored in file) via method.
-        supported dictionary methods are: update(default), pop, clear
-        and data is attribute of the method
-
-        Modification will invalidate affected node data and childs, please reload the Tree
-
-        Be aware that it modify raw data in file:
-        If you have defined "key+: fist" in the file for node and
-        you will add "key: second" it will lead  to "secondfirst" output for node
-
-        Data are always stored into the last sourced fmf file of the modified node
-          (e.g no inheritance and no elasticity)
-        """
-        node_data = self._get_node_raw_data(default=dict())
-        if method in ["update", "pop"]:
-            for item in data:
-                getattr(node_data, method)(item)
-        elif method == "clear":
-            if data:
-                raise AttributeError("Unexpected argument for clear method", data)
-            else:
-                node_data.clear()
-        else:
-            raise ValueError("Unsupported method: {} (you can use dictionary methods: update, pop, clear)".format(method))
-        return self
-
-    def save(self):
-        """
-        Store data back to fmf files after modification
-        """
-        raw_data_node, _ = self._get_raw_data_node()
-        target_file = raw_data_node.sources[-1]
-        with open(target_file, "w") as fd:
-            fd.write(dict_to_yaml(raw_data_node._raw_data))
+        # The full raw data were read from the last source
+        return node_data, full_data, node.sources[-1]
 
     def __enter__(self):
-        return self._get_node_raw_data(default=dict())
+        """
+        Experimental: Modify metadata and store changes to disk
+
+        This provides an experimental support for storing modified node
+        data to disk. For now, the implementation is very simple, data
+        are always stored into the last source file from which node data
+        were read.
+
+        The provided object contains only raw data. There is no support
+        for inheritance, elasticity or data merging. For example, if you
+        have defined "key+: value" in the file for node and you will add
+        "key: other" it will result into "othervalue".
+
+        Example usage:
+
+            with Tree('.').find('/tests/core/smoke') as test:
+                test['tier'] = 0
+
+        Note that white space will be stripped and comments removed as
+        export to yaml does not preserve this information. The feature
+        is experimental and can be later modified, use at your own risk.
+        """
+        return self._locate_raw_data()[0]
 
     def __exit__(self, exc_type, exc_val, exc_tb):
-        self.save()
+        """ Experimental: Store modified metadata to disk """
+        _, full_data, source = self._locate_raw_data()
+        with open(source, "w") as file:
+            file.write(dict_to_yaml(full_data))
--- a/fmf/base.py
+++ b/fmf/base.py
@@ -14,7 +14,7 @@ from filelock import Timeout, FileLock
 import fmf.context
 import fmf.utils as utils
 from io import open
-from fmf.utils import log
+from fmf.utils import log, dict_to_yaml
 from pprint import pformat as pretty
 
 # ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
@@ -73,6 +73,7 @@ class Tree(object):
         self.version = utils.VERSION
         self.original_data = dict()
         self._commit = None
+        self._raw_data = dict()
 
         # Special handling for top parent
         if self.parent is None:
@@ -400,6 +401,7 @@ class Tree(object):
         # Save source file
         if source is not None:
             self.children[name].sources.append(source)
+            self.children[name]._raw_data = data
 
     def grow(self, path):
         """
@@ -446,6 +448,7 @@ class Tree(object):
             # Handle main.fmf as data for self
             if filename == MAIN:
                 self.sources.append(fullpath)
+                self._raw_data = data
                 self.update(data)
             # Handle other *.fmf files as children
             else:
@@ -606,3 +609,77 @@ class Tree(object):
         duplicate = copy.deepcopy(self)
         self.parent = duplicate.parent = original_parent
         return duplicate
+
+    def _get_raw_data_node(self):
+        """
+        Get tuple
+          - node where _raw_data are stored
+          - list if keys for current node inside raw data
+        """
+        raw_data_node = self
+        id_list = list()
+        while True:
+            raw_data = raw_data_node._raw_data
+            if raw_data:
+                return raw_data_node, id_list
+            elif not raw_data_node.parent:
+                raise utils.RootError("Cannot find root node with _raw_data, very strange")
+            else:
+                id_list.insert(0, "/" + raw_data_node.name.rsplit("/")[-1])
+                raw_data_node = raw_data_node.parent
+
+    def _get_node_raw_data(self, default=None):
+        """
+        Get dictionary of _raw_data for current node
+        in case of None add to the current node key default element
+        """
+        raw_data_node, id_list = self._get_raw_data_node()
+        raw_data = raw_data_node._raw_data
+        for key in id_list:
+            if default is not None and raw_data[key] is None:
+                raw_data[key] = default
+            raw_data = raw_data[key]
+        return raw_data
+
+    def modify(self, *data, method="update"):
+        """
+        Modify node metadata (original data stored in file) via method.
+        supported dictionary methods are: update(default), pop, clear
+        and data is attribute of the method
+
+        Modification will invalidate affected node data and childs, please reload the Tree
+
+        Be aware that it modify raw data in file:
+        If you have defined "key+: fist" in the file for node and
+        you will add "key: second" it will lead  to "secondfirst" output for node
+
+        Data are always stored into the last sourced fmf file of the modified node
+          (e.g no inheritance and no elasticity)
+        """
+        node_data = self._get_node_raw_data(default=dict())
+        if method in ["update", "pop"]:
+            for item in data:
+                getattr(node_data, method)(item)
+        elif method == "clear":
+            if data:
+                raise AttributeError("Unexpected argument for clear method", data)
+            else:
+                node_data.clear()
+        else:
+            raise ValueError("Unsupported method: {} (you can use dictionary methods: update, pop, clear)".format(method))
+        return self
+
+    def save(self):
+        """
+        Store data back to fmf files after modification
+        """
+        raw_data_node, _ = self._get_raw_data_node()
+        target_file = raw_data_node.sources[-1]
+        with open(target_file, "w") as fd:
+            fd.write(dict_to_yaml(raw_data_node._raw_data))
+
+    def __enter__(self):
+        return self._get_node_raw_data(default=dict())
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self.save()
--- a/tests/unit/test_modify.py
+++ b/tests/unit/test_modify.py
@@ -3,26 +3,22 @@
 from __future__ import unicode_literals, absolute_import
 
 import unittest
+import pytest
 import os
+import re
 import tempfile
-from fmf.base import Tree
-from shutil import rmtree, copytree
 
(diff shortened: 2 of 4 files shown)


lukaszachy:
Looks good, for simple key, but +- suffixes can break it.
1. Create main.fmf
```
demo: a
/first:
    demo+: b
```
2. Verify it works `fmf show`
```
/first
demo: ab
```
3. Load in fmf and change it
```python
import fmf
tree = fmf.Tree('.')
node = list(tree.prune())[0]
node.modify({'demo':'nope'})
```
4. Check main.fmf
```
demo: a
/first:
    demo+: b
    demo: nope
```
5. And see it is different `fmf show`
```
/first
demo: nopeb
```

jscotka:
@lukaszachy it is question if it breaks it. 
From perspective of FMF it is not  a bug and it is proper syntax.

from the perspective of various ways of implementation there are more ways what could be intuitive for different people:
 1. current solution - it also makes sense. it exactly does what you write
   ```
demo: a
/first:
    demo+: b
    demo: nope
```
 2. it preserve original merging item, and just replace the item with value
   ```
demo: a
/first:
    demo+: nope
```
 3. it removes all other merging items and put there just value what you've said.
   ```
demo: a
/first:
    demo: nope
```
 4. `raise Error`, that there already exist item, and you have to remove `item-/+` first
 5. then exists also solutions what will respect the value inside other +- items and will apply merging ways on original.
 6. allow to use some parameters or various `modify` method names what will impement previous possible solutions.

And also look at it from perspective you would like to do: `node.modify({'demo+':'nope'})` does there apply same rules or it will change your PoV how to deal with solution?

I'm not against any solution and any of solution makes sense and it is not problem to implement it.
My point is just about that there does not exists right solution, just solution what currently the most suits your PoV.

File: fmf/base.py [hunk 1]
```diff
@@ -544,3 +547,58 @@ def node(reference):
                 "No tree node found for '{0}' reference".format(reference))
         # FIXME Should be able to remove .cache if required
         return found_node
+
+    def get_raw_data_node(self):
```
lukaszachy, line 551: Could this be a private method? Personally I don't see why user of fmf library would call this.
However if there is need for it being public docstring will be very nice to have

fmf/base.py
```diff
@@ -544,3 +547,58 @@ def node(reference):
...
+            raw_data = raw_data_node.raw_data
+            if raw_data:
+                return raw_data_node, id_list
+            elif not raw_data_node.parent:
+                raise utils.RootError("Cannot find root node with raw_data, very strange")
+            else:
+                id_list.insert(0, "/" + raw_data_node.name.rsplit("/")[-1])
+                raw_data_node = raw_data_node.parent
+
+    def node_get_raw_data(self):
```
lukaszachy, line 564: Could this be a private method?

fmf/base.py
```diff
@@ -544,3 +547,58 @@ def node(reference):
...
+        raw_data_node, id_list = self.get_raw_data_node()
+        raw_data = raw_data_node.raw_data
+        for key in id_list:
+            raw_data = raw_data[key]
+        return raw_data
+
+    def modify(self, data, method="update", reload=False):
+        """
+        Modify node metadata (original data stored in file) via method.
+        method is any method what contains python dictionary (default: update)
```
lukaszachy, line 574: Maybe an explicit note that it updates (saves) the file.

fmf/base.py
jscotka: I agree, probably nobody just FMF hackers would use this method, so private is fine here.

fmf/base.py
```diff
@@ -544,3 +547,58 @@ def node(reference):
...
+        if reload:
+            self.reload()
+        return self
+
+    def __data_cleanup(self):
+        for item in self.climb(whole=True):
+            item.data = dict()
+            item.sources = []
+
+    def reload(self):
```
lukaszachy, line 593: Please add a docstring as it can be used from the outside

fmf/base.py
jscotka: This method coud be useful. in case user wants to modify data in theirs way and do not rely on modify.update.
I can document it here. as we've discussed various ways of modifiying current node.
But theoretically it could be also private, because, probably just me and you understand why it is good for. and in case user want another modify mechanism could create new issues and dissuss way how and why.

fmf/base.py
```diff
@@ -544,3 +547,58 @@ def node(reference):
...
+        source_file = raw_node.sources[-1]
+        raw_node.__data_cleanup()
+        raw_node.grow(os.path.dirname(source_file))
+        raw_node.inherit()
+
+    def store(self):
+        raw_data_node, _ = self.get_raw_data_node()
+        target_file = raw_data_node.sources[-1]
+        with open(target_file, "w") as fd:
+            fd.write(dict_to_yaml(raw_data_node.raw_data))
```
lukaszachy, line 604: This will need a lock, but I'll  address that in #97

jscotka: yep, locking of modification of files is another issue

tests/unit/test_modify.py
```diff
@@ -0,0 +1,111 @@
+# coding: utf-8
+
+from __future__ import unicode_literals, absolute_import
+
+import unittest
+import os
+import tempfile
```
lukaszachy, line 7: Since the testsuite is using pytest one can use https://docs.pytest.org/en/stable/tmpdir.html#the-tmpdir-fixture
However it should be OK to keep as it is now.

lukaszachy (changes requested):
Overall looks good, some methods should be private and public once need docstrings

fmf/base.py
```diff
@@ -548,7 +548,11 @@ def node(reference):
         # FIXME Should be able to remove .cache if required
         return found_node
 
-    def get_raw_data_node(self):
+    def __get_raw_data_node(self):
+        """
+        Get tuple of node where raw_data are stored and list if keys
+        for current node inside raw data
```
lukaszachy, line 582: I don't understand this comment

lukaszachy (approved):
One dosctring (of private method) is not clear to me but LGTM otherwise

fmf/base.py
jscotka: there is typo: "list if keys" -> "list of keys"

psss:
Thanks a lot for looking into this, @jscotka! I was experimenting a bit with the `reload` functionality:

```python
import fmf
import pprint

print('\nBefore:')
deep = fmf.Tree('.').find('/recursion/deep')
pprint.pprint(deep.data)

print('\nAfter:')
deep.modify(dict(test='./test.sh', x=1), reload=True)
pprint.pprint(deep.data)
```

And here's the result:

```
Before:
{'depth': 1000,
 'description': 'Check recursive download options',
 'path': 'wget/recursion',
 'tags': ['Tier2'],
 'test': 'runtest.sh',
 'tester': 'Petr Šplíchal <psplicha@redhat.com>',
 'time': '10 min'}

After:
{'description': 'Check recursive download options',
 'path': 'wget/recursion',
 'tags': ['Tier2'],
 'test': 'runtest.sh',
 'tester': 'Petr Šplíchal <psplicha@redhat.com>'}
```

The new attribute `x` is not added, existing key `test` is not updated and attributes `depth` and `time` disappear. Am I doing something wrong?

fmf/base.py
psss: Plus one for better clarifying the method meaning.

fmf/base.py
```diff
@@ -588,3 +590,80 @@ def copy(self):
         duplicate = copy.deepcopy(self)
         self.parent = duplicate.parent = original_parent
         return duplicate
+
+    def __get_raw_data_node(self):
```
psss, line 594: All internal methods use a single leading underscore, let's keep this consistent.

fmf/base.py
```diff
@@ -588,3 +590,80 @@ def copy(self):
...
+        Or reload whole tree to ensure data are completely fine.
+        """
+
+        getattr(self.__get_node_raw_data(), method)(data)
+        self.__store()
+        if reload:
+            self.reload()
+        return self
+
+    def __data_cleanup(self):
```
psss, line 640: Please, add a docstring.

fmf/base.py
```diff
@@ -588,3 +590,80 @@ def copy(self):
...
+
+    def reload(self):
+        """
+        Try to reload data after it is modified
+        It reloads data from node where are defined.
+
+        WARNING:
+        This method may lead that some variables to nodes in your code could lost references.
+        In case of bigger modification, it would be better to reload whole tree
+        in your code.
```
psss, line 653: Well, I'm not completely sure if we should provide a functionality which works only maybe or sometimes... If this is too complex, shall we leave it for later and if really needed? Anyway, why is `reload` needed?

psss (changes requested):

psss:
One more simple example in the `scatter` directory:

```python
import fmf
import pprint

print('\nBefore:')
scatter = fmf.Tree('.').find('/object')
pprint.pprint(scatter.data)

print('\nAfter:')
copy = scatter.modify(dict(one=11, two=22, three=33), reload=True)
pprint.pprint(scatter.data)
```

Works good but modified only a single source file:

```
> python test.py 

Before:
{'one': 1, 'three': 3, 'two': 2}

After:
{'one': 11, 'three': 33, 'two': 22}

> git diff
diff --git a/examples/scatter/object/main.fmf b/examples/scatter/object/main.fmf
index 4f569db..258a172 100644
--- a/examples/scatter/object/main.fmf
+++ b/examples/scatter/object/main.fmf
@@ -1 +1,3 @@
-three: 3
+three: 33
+one: 11
+two: 22
```

fmf/base.py
```diff
@@ -588,3 +590,60 @@ def copy(self):
...
+
+        Be aware that it modify raw data in file:
+        If you have defined "key+: fist" in the file for node and
+        you will add "key: second" it will lead  to "secondfirst" output for node
+
+        Data are always stored into the last sourced fmf file of the modified node
+          (e.g no inheritance and no elasticity)
+        """
+
+        getattr(self._get_node_raw_data(), method)(data)
```
psss, line 638: I've run into the following traceback:

```python
>>> fmf.Tree('examples/wget').find('/download/requirements/spider').modify(dict(x=1))
  File "<stdin>", line 1, in <module>
  File "/home/psss/git/fmf/fmf/base.py", line 636, in modify
    getattr(self._get_node_raw_data(), method)(data)
AttributeError: 'NoneType' object has no attribute 'update'
```

Are you able to reproduce as well?

jscotka: Good catch @psss it is caused by no data inside the dict element. I've expected to be empty dict, but it is None.   So that I have to init dict there.

jscotka:
regarding to discussion, it may also behave as decorator. I can imagine also some code like
```
node.modify(update_dict)
node.modify("pop")("item_to_remove")
```
or some more clean solution, like modify/create child `builtin` `dict` class, what will implement the expected behaviour.
so that `self.raw_data` will be class. So that there will be no modify method, but will there will `self.raw_data` or `self.modify` `property` 
and there will be class like:
```
class OwnDict:
    def __init__(value, file):
      self.dict = value
      self.file = file
    def update(dict1)
      # what currently does modify and store function
    def pop(item_list)
...
```
what do you think?

psss:
I'd say it's getting a bit complicated. I don't think user should know about `raw_data` at all. What about something simple like this?

```python
if method == 'update':
    raw_data.update(data)
elif method == 'pop':
    raw_data.pop(data)
elif method == 'clear':
    raw_data.clear()
else:
   raise GeneralError("Unsupported method...")
```

And explicitly state we support these three methods?

jscotka:
> I'd say it's getting a bit complicated. I don't think user should know about `raw_data` at all. What about something simple like this?
> 
> ```python
> if method == 'update':
>     raw_data.update(data)
> elif method == 'pop':
>     raw_data.pop(data)
> elif method == 'clear':
>     raw_data.clear()
> else:
>    cccccccc
> ```
> 
> And explicitly state we support these three methods?

Sure, I agree, for me it is clear, and mabe do that as `*data` in definition and maybe then use someething like
```
def modify(*data, method="update")
  if method in ["update", "pop"]:
     for item in data:
       getattr(self.raw_data, method)(item)
  elif method == "clear":
     if data:
        raise AttributeError("Unexpected argument for clear method", data)
     else:
        self.raw_data.clear()
  else:
    raise GeneralError("Unsupported method...")
```

fmf/base.py
```diff
@@ -641,13 +641,13 @@ def _get_node_raw_data(self, default=None):
             raw_data = raw_data[key]
         return raw_data
 
-    def modify(self, data, method="update"):
+    def modify(self, *data, method="update"):
```
psss, line 644: Looks fine. Just not sure whether it's better to use `data` or `*data` here. The second has advantage that one could provide multiple dictionaries, not sure how frequent such use case would be. Any other pros? @lukaszachy, any preference here?

psss:
One more idea for discussion. What about this?

    Tree.set() .... for modifying node data (as a sister method for Tree.get())
    Tree.save() ... for storing data to disk?

Would something like that make sense? One could do several node data adjustments and then store the whole change.

fmf/base.py
jscotka: I can see here these pros and cons:
- data
  - pros
    - update - cleaner it takes exactly one arg
    - pop - pop also takes righ one arg
  - cons
    - cannot pop more values in one command, what i see useful
    - clear command is strange that you have to set value, or put there some default like None, but it may lead to situation that you have to chech value for update or pop

- *data
  - pros:
    - pop - easy to remove more items
    - clear - no defaut value needed and expect no additional value
  - cons:
    - update - little bit strange possibility to update same dict moretimes,is it useful, probably not much from my PoV

jscotka:
> One more idea for discussion. What about this?
> 
> ```
> Tree.set() .... for modifying node data (as a sister method for Tree.get())
> Tree.save() ... for storing data to disk?
> ```
> 
> Would something like that make sense? One could do several node data adjustments and then store the whole change.

I've also though about this, to explicitly call 'save' but i can see to be easy to forgot, but i can imagine more useful solutions, how to do it.
1. Define __enter/__exit methods and do it then with "with" statement and store when leaving block
2. Define __del__ method to ensure to store data as fallback, this is nice but could be not so reliable.
3.  Your explicit set/store is also possible, but not so user friendly from my PoV

fmf/base.py
lukaszachy: Personally I like the approach of e.g. [Thread](https://docs.python.org/3/library/threading.html#threading.Thread) - `target=None, args=(), kwargs={}`. Just rename target to method so it is clear that it is name of the dict method as a string.

jscotka: > Personally I like the approach of e.g. [Thread](https://docs.python.org/3/library/threading.html#threading.Thread) - `target=None, args=(), kwargs={}`. Just rename target to method so it is clear that it is name of the dict method as a string.

Thats also not bad, but what I don't like is the syntax then, so that, instead of using `node.modify({a=b, c=d})` you have to use `node.modify(args=({a=b, c=d},))` (yes, you can also have there possibility to check if args is not iterable, use it as one item, but it is more tricky from type annotation then) what is confusing for me. but it is cleaner from perspective, that args, and kwargs are posted to `method`
In fact some tools using this and another uses `*args, **kwargs` and move these explicitly undefined attributes into internally used methods.

psss:
> I've also though about this, to explicitly call 'save' but i can see to be easy to forgot, but i can imagine more useful solutions, how to do it.

I don't think "remember to save to disk" should present a significant problem. I like the explicit nature of the `save()` method. Also I consider it a valid use case to work the the metadata trees in memory without saving every change to disk. For example you could do something like this:

    node.set('key', 'value')
    node.pop('obsolete')
    node.save()

In the future we could extend the `save()` method to support for example saving all children as well.

jscotka:
> > I've also though about this, to explicitly call 'save' but i can see to be easy to forgot, but i can imagine more useful solutions, how to do it.
> 
> I don't think "remember to save to disk" should present a significant problem. I like the explicit nature of the `save()` method. Also I consider it a valid use case to work the the metadata trees in memory without saving every change to disk. For example you could do something like this:
> 
> ```
> node.set('key', 'value')
> node.pop('obsolete')
> node.save()
> ```
> 
> In the future we could extend the `save()` method to support for example saving all children as well.

@psss okay, makes sense, explicit save needed, but thanks to returning `self`, you can also chain operations as well so possible to do:

```python
node.modify(dict(key='value'))
node.modify('obsolete', method="pop")
node.save()
```

```python
node.modify(dict(key='value')).modify('obsolete', method="pop").save()
```
or add also `with` statement to store data back to disc
```python
with Tree(root).find("/node/to/modify") as data:
    data.update(dict(key='value'))
    data.pop('obsolete')
```
Using this with statement is much clearer, because it exactly returns exact part of dictionary and you can directly use the dict methods, or implement logic what you want, e.g. remove all other keys
```python
with Tree(root).find("/node/to/modify") as data:
    new_key = "key"
    # remove - + items if they are there
    for key in data:
        if new_key in key:
            data.pop(key) 
    data[new_key]='value'
```




is it clearer now?

jscotka:
@psss @lukaszachy please, please  merge :-)

lukaszachy (approved):
No more comments, LGTM

psss:
After discussion with @jscotka I've removed the `modify()` method as the context manager solution nicely and in much better way covers all use cases related to data dictionary modification. In order to improve readability I've merged the two internal methods for locating raw data into one and added a couple of comments explaining how the magic works. Tests adjusted accordingly. Please, review.

psss (approved):

psss:
Multi-line strings should be fixed now.

jscotka (commented):
LGTM. I can add just comment, not able to add approval :-)

lukaszachy:
Hm, should this work?
fmf root:
```
$ ls -1 | xargs head 
==> child.fmf <==
foo: value
==> main.fmf <==
long: |
  first line
  second line
  third line
```
Trying to modify:
```python
In [4]: with fmf.Tree('/tmp/z').find('/child') as item:
   ...:     item['long'] = item['long'] + '\nnext line'
   ...:      
   ...:                                                                                                                                                                              
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
<ipython-input-4-cccea106c5bc> in <module>
      1 with fmf.Tree('/tmp/z').find('/child') as item:
----> 2     item['long'] = item['long'] + '\nnext line'
      3
      4
 
KeyError: 'long'
> <ipython-input-4-cccea106c5bc>(2)<module>()
      1 with fmf.Tree('/tmp/z').find('/child') as item:
----> 2     item['long'] = item['long'] + '\nnext line'
      3
      4
In [5]: fmf.Tree('/tmp/z').find('/child').get('long')                                                                                                                                
Out[5]: 'first line\nsecond line\nthird line\n'
```

jscotka:
> ```python
> em['long']
> ```

it is exactly trouble what is described there. there is no merging, elasticity or scattering feature, and item['long'] is not defined inside the child item, and you try to find it. it is defined in upper item, in `main.fmf`,  so that there could work something like:

```
item['long'] = fmf.Tree('/tmp/z').find('/child').data["long"] + "new line"
```

or 
```
item['long+'] = "new line"
```

psss:
Yes, as @jscotka points out, this is the limitation of this initial experimental implementation. In the future, I can imagine we would be able to do some magic so that we're able to locate even inherited metadata but for now I'd say it is good as it is.

jscotka:
> Yes, as @jscotka points out, this is the limitation of this initial experimental implementation. In the future, I can imagine we would be able to do some magic so that we're able to locate even inherited metadata but for now I'd say it is good as it is.

I have a fear that it is not possible to improve the process there because there are so many ways that could be possible on the data level.

But simplest solution is possible via some wrappers what will help you find the proper node in hierarchy for your purposes (e.g. your case) or similar could be find any item (including +- items): 
```
def find_node_where_defined(node, key):
    current, _, _ = _locate_raw_data(node)
    if key in current:
        return node
    if node.parent:
        return find_node_where_defined(node.parent, key)
```

psss:
Yes, there is a space for improvement (although the implementation will not be quite simple I guess) but let's unblock the `tmt test export` first with the basic functionality.

lukaszachy (approved):
LGTM with what is possible to do today

psss:
Thanks for the review. Squashed and merged.

PR #104, opened 2021-02-04
Use the origin default branch instead of master

psss:
Instead of `master` let's use the default branch from the origin
if `ref` is not provided when fetching remote repositories. Plus
a minor update of the run() function to improve debugging.

### DIFF
--- a/fmf/base.py
+++ b/fmf/base.py
@@ -545,7 +545,7 @@ class Tree(object):
         Keys supported in the reference:
 
         url .... git repository url (optional)
-        ref .... branch, tag or commit ('master' by default)
+        ref .... branch, tag or commit (default branch if not provided)
         path ... metadata tree root ('.' by default)
         name ... tree node name ('/' by default)
 
@@ -558,7 +558,7 @@ class Tree(object):
         if 'url' in reference:
             path = reference.get('path', '.').lstrip('/')
             repository = utils.fetch(
-                reference.get('url'), reference.get('ref', 'master'))
+                reference.get('url'), reference.get('ref'))
             root = os.path.join(repository, path)
         # Use local files
         else:
--- a/fmf/utils.py
+++ b/fmf/utils.py
@@ -9,6 +9,7 @@ import re
 import sys
 import copy
 import time
+import shutil
 import logging
 import subprocess
 from pprint import pformat as pretty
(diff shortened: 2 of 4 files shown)


fmf/base.py
```diff
@@ -517,7 +517,7 @@ def node(reference):
         Keys supported in the reference:
 
         url .... git repository url (optional)
-        ref .... branch, tag or commit ('master' by default)
+        ref .... branch, tag or commit (default branch by default)
```
lukaszachy, line 520:
```suggestion
        ref .... branch, tag or commit (default branch if not present)
```

lukaszachy: IMO otherwise can be read that 'default branch' is a possible value. Or at least for me:)

fmf/utils.py
```diff
@@ -532,11 +534,19 @@ def fetch(url, ref='master'):
     directory = url.replace('/', '_')
     destination = os.path.join(fmf_cache, directory)
 
+    # Use the default branch from origin of no ref provided
+    if ref is None:
+        ref = 'DEFAULT'
```
lukaszachy, line 539:
```suggestion
        ref = '__DEFAULT__'
```

lukaszachy: Not sure how often one has branch DEFAULT but __DEFAULT__ seems to be more rare.

fmf/utils.py
```diff
@@ -532,11 +534,19 @@ def fetch(url, ref='master'):
...
+
     try:
         # FIXME Implement locking
         # Clone the repository
         if not os.path.isdir(destination):
             run(['git', 'clone', url, destination], cwd=fmf_cache)
+            # Store the default branch from the origin as a DEFAULT ref
+            head = os.path.join(destination, '.git/refs/remotes/origin/HEAD')
+            default = os.path.join(destination, '.git/refs/heads/DEFAULT')
```
lukaszachy, line 548:
```suggestion
            default = os.path.join(destination, '.git/refs/heads/__DEFAULT__')
```

tests/unit/test_utils.py
```diff
@@ -155,6 +156,16 @@ def test_color(self):
 class TestFetch(object):
     """ Remote reference from fmf github """
 
+    def test_fetch_default_branch(self):
```
lukaszachy, line 159: Could you add a test which changes branch back to default branch?
1. fetch node from explicit, no-default branch
2. fetch node from default branch (without specifying ref:)

lukaszachy: Also extending `test_out_of_sync_ref` with ref=None will be nice

fmf/base.py
psss: I see, `provided` sounds good as well?

fmf/utils.py
psss: Yeah, let's minimize possible conflict.

tests/unit/test_utils.py
psss: I've added a new test `test_switch_branches` to cover branch switching. Regarding `test_out_of_sync_ref`, here the default branch probably does not make sense, because `__DEFAULT__` is a ref to `origin/master` and thus it cannot be out of sync from it. If the fresh `origin/master` is fetched when cache expires `__DEFAULT__` will be still correct.

lukaszachy (approved):
Thank you, LGTM

PR #106, opened 2021-02-10
Support calls fetch(destination, env) and run(env)

lukaszachy:
Related to https://github.com/psss/tmt/issues/530#issuecomment-775781414


lukaszachy:
Hm, we will need 'env' as well, necessary to be used as https://github.com/psss/tmt/blob/3e4673b5955166f880c19608a3bdf902661bbf01/tmt/beakerlib.py#L129

psss:
Hooray, everything's green! :)

PR #107, opened 2021-02-12
Enable integration testing with tmt, some cleanup

psss:
Create a new integration plan for tmt.
Remove old execute step methods.
Update .gitignore list.


psss:
All tests passed except for rawhide which has a problem with gpg keys.

PR #112, opened 2021-03-05
Add validate_fmf_id function

thrix:
Introduced to `tmt`, but seems useful directly in `fmf` library.

Signed-off-by: Miroslav Vadkerti <mvadkert@redhat.com>

### DIFF
--- a/fmf/utils.py
+++ b/fmf/utils.py
@@ -12,6 +12,7 @@ import time
 import shutil
 import logging
 import subprocess
+import collections
 from filelock import Timeout, FileLock
 from pprint import pformat as pretty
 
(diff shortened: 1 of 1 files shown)


psss:
I find the mapping implementation a bit cryptic. It also requires all fmf id dictionary keys to be defined. I think several if/elif statements would make the code much more readable. Also one brainstorming idea: What about throwing exceptions upon errors and returning just `True` when everything is fine? Seems more natural than `return (True, '')`. Also this could be directly handled in `Tree.node()` and this extra function would not be necessary.

thrix:
we decided to drop this

Issue #77, opened 2020-04-03
Add support for cache cleanup


psss:
The `Tree.node()` method fetches fmf metadata into `~/.cache/fmf`. This can grow in time. It would be nice to implement an easy way to clean up old/all cached stuff. There should be available both from python API and command line. Brainstorming command line usage:

    fmf gc
    fmf gc --all
    fmf gc --old

lukaszachy:
To elaborate a bit,  As a user I want **fmf** to
- clean after itself so my home/.cache stays clean
- clean old stuff so I free some space on disk
- fetch current remote so latest changes are present in metadata repository

At the moment we do `git fetch` if last fetch was more than 20 minutes ago but it is not configurable.

So how about adding option instead of command? To be a bit inspired by dnf: there is `--refresh` which forces metadata refresh and command 'dnf clean'

```
fmf --clean [time] [command] 
fmf --refresh [command]
```

So if you don't want to keep cache, you use `--clean`, `fmf --clean` will erase existing.
To keep just recent cache `fmf --clean 1d` if you recognize 1 day as recent ;)
Refresh all on `fmf --refresh`

lukaszachy:
@psss Thoughts? I'd try to implement it if you are OK with the above.

psss:
A couple of questions:

* Would the `--clean` option erase the cache before doing the command or after? With a separate command one could do both and it would be clear when the cleanup happens.
* Anything cached longer than 20 minutes is currently fetched again, so it probably does not make sense to only clean up items older than 1d, anything between 1d and 2m will not be (re)used anyway

So far the command variant seems to me as a better solution but I might be overlooking possible advantages of the option solution.

lukaszachy:
> So far the command variant seems to me as a better solution but I might be overlooking possible advantages of the option solution.

How would you do, with single command, following?
1. As a user I want fmf to ignore cache and fetch repositories directly
UC: I've just pushed changes to the upstream
`fmf --refresh [commnad]`
2. As a user I want fmf to clean after itself so no files are left in .cache.
`fmf --clean [commnad]`
UC: I'm on shared machine / don't have much space left
3. As a user I want fmf to ignore any cache (might not be there) and clean after itself.
`fmf --refresh --clean [command]`

I prefer single command as it is easier to 'alias' or recall from bash history or even copy&paste from help or IRC

Good point that `--clean [time]` doesn't make sense.

PR #114, opened 2021-03-12
Implement 'gc' command

lukaszachy:
Solves #77

In the end I went with simple gc command to delete whole cache dir.
However API allows to set custom expiration time.


lukaszachy:
Renamed command to 'clean' and all methods it uses are part of their own PR  #118  so it is easier for review / changes.

lukaszachy:
Found issue in api-cache branch (base of this PR) so rebasing.

psss (approved):
Nice, looks good and works as expected. Thanks!

PR #115, opened 2021-03-17
Fetch whole tree using utils.get_remote_tree

lukaszachy:
Moved fetching of the tree to utils, closer to where it belongs. Tree.node just uses this function,

Main main goal is to use this function from wow (and cache results) so fmf seems a better place to take care of locking (as should operate on fmf cache).

### DIFF
--- a/fmf/utils.py
+++ b/fmf/utils.py
@@ -16,6 +16,9 @@ from filelock import Timeout, FileLock
 from pprint import pformat as pretty
 import io
 import yaml
+import warnings
+
+import fmf.base
 
 
 # ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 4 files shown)


fmf/utils.py
```diff
@@ -565,6 +569,32 @@ def set_cache_directory(cache_directory):
     global _CACHE_DIRECTORY
     _CACHE_DIRECTORY = cache_directory
 
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+# Get Whole Tree from remote
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+def get_remote_tree(url, ref, path):
```
lukaszachy, line 575: I'm not sure about default values - I'd prefer to have just one place where they belong - fmf.utils.fetch in this case.

psss:
Yeah, makes sense to keep both fetching the repo and the tree under `utils`. Thinking about the name... to make it consistent... as this is related to fetching remote objects, what about calling it `fetch_tree()`? Would be better if the other one could be `fetch_repo()` to make the distinction clear (and keep `fetch()` backward-compatibility for some time). What do you think?

psss (approved):
Looks great, merging with just a few cosmetic changes. Thanks!

PR #120, opened 2021-03-24
A bunch of adjustments for the epel-7 build

psss:
Enable the `epel-7` target in packit.
Build only the `python2-fmf` package for rhel-7.
Resolve several Python 2 incompatibilities.
Enable unit tests, correctly mark all web tests.
Skip test which needs pytest.warns module.
Skip selected tests when run as root.
Adjust requires accordingly, add missing `git-core`.

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -18,6 +18,7 @@ jobs:
       targets:
       - fedora-all
       - epel-8
+      - epel-7
 
   - job: copr_build
     trigger: commit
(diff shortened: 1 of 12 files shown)


lukaszachy:
@psss I'm not sure how to disable `"/plans/integration"` for rhel-7 only. 
The rest passes however I made them to be skipped when they run as root or when pytest is too old respectively.
I think we could make permission denied to be mocked, but I'd leave that for next time.

lukaszachy (approved):

psss:
@lukaszachy, thanks a lot for working on this together!

PR #121, opened 2021-03-26
Do not use mutable objects as default arguments

FrNecas:
Using lists/dicts as default arguments is a bad practice that can lead
to unexpected errors with subsequent calls to the method (the same
default instance is shared across all subsequent calls).

### DIFF
--- a/fmf/base.py
+++ b/fmf/base.py
@@ -502,8 +502,14 @@ class Tree(object):
                 return node
         return None
 
-    def prune(self, whole=False, keys=[], names=[], filters=[], conditions=[]):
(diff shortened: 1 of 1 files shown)


jscotka (approved):
seems nice, thanks a lot.

lukaszachy (approved):
LGTM

lukaszachy:
Let's wait with the merge until #120 is ready (it adds rhel-7 tests)

psss:
Thanks for the fix. Let's squeeze this into `0.15.2` if tests pass after rebase on the latest master.

PR #124, opened 2021-03-30
Use a deep copy when storing raw node data

psss:
When the `adjust()` method is used it can modify data referenced
from the raw data dictionaries. Make a deep copy to ensure the
original data are preserved.

- [ ] add test coverage

### DIFF
--- a/fmf/base.py
+++ b/fmf/base.py
@@ -318,7 +318,7 @@ class Tree(object):
 
         # Adjust rules should be a dictionary or a list of dictionaries
         try:
-            rules = self.data.pop(key)
+            rules = copy.deepcopy(self.data[key])
             log.debug("Applying adjust rules for '{}'.".format(self))
             log.data(rules)
             if isinstance(rules, dict):
--- a/fmf/base.py
+++ b/fmf/base.py
@@ -419,7 +419,7 @@ class Tree(object):
         # Save source file
         if source is not None:
             self.children[name].sources.append(source)
-            self.children[name]._raw_data = data
+            self.children[name]._raw_data = copy.deepcopy(data)
 
     def grow(self, path):
         """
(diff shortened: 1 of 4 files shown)


jscotka (approved):
Hi, seems well.
I've also met this issue, but though that issue is on my side with regerated data from python decorators.
yep, seems like `adjust()` causes it. BTW, would not be better to fix adjust method if it somehow modify something what it should not modify? e.g. there can be found another issue with method use parameter and reference and potential future problem. and also reason why I've added lines for adjust to test https://github.com/psss/fmf/blob/master/tests/unit/test_modify.py#L82 :-) to prove that it does not modify adjust rule.

lukaszachy:
I believe it would be better to allow repetitive `adjust`  calls - tmt object stays "adjusted" until next adjust call. 
AFAIK Alex and I need to remember tmt state before adjust (`copy` method)  as our tooling need to call adjust multiple times.

jscotka:
my feeling is that it should be keep. I can imagine that in pipeline some additional step may change adjust output, e.g. some artificial example :-) I can have centOS and will use RH tool for converting centos to RHEL, and then adjust have to match that it is not centos but rhel (e.g. in `prepare` phase). I understand that it is little bit artificial, but way, that multiple call of adjustment is not bad. but feature. from py PoV

psss:
Makes sense. Let's allow repeated `adjust()` calls then. Will update the code.

jscotka:
@psss @lukaszachy maybe as quick fix we can also use this patch and create new one (or issue) for your mentioned solution. Because it will be then also feature not just bugfix.

lukaszachy (approved):
I've added test, @psss please review.

psss:
Thanks, @lukaszachy. The test looks good. Added changes needed to keep the original adjust rules intact, including a simple test.

lukaszachy:
Hm, so integration fails on f34 as tmt changed `'enabled\s+yes'` to `true`

psss:
Yeap, another motivation to find an elegant solution for psss/tmt/issues/585 as soon as possible.

PR #133, opened 2021-04-14
Add support for empty files and virtual nodes

### LINKED ISSUE #131: Empty fmf file should mark that there exist node
I have ucasese, that I've expected that also empty fmf file creates FMF node, but I've found that it is not true
like:
```
mkdir x
cd x
fmf init
touch main.fmf
fmf ls
```
or more complex:
```
 echo a: b > main.fmf
touch a.fmf
touch b.fmf
mkdir c
touch c/main.fmf
```

or also way, to so simple inheritance in one file `main.fmf`:
```
a: b
/c: {}
/d: {}
```

So there is no way how to simply create inherited nodes, without any data in fmf files.
But I think this way should exists.

psss:
Empty files and virtual nodes are no more ignored.
The 'fmf show' command now show empty nodes as well.
Test coverage adjusted, new test covering cli added.

Resolves #131.


fmf/base.py
```diff
@@ -103,11 +106,11 @@ def __init__(self, data, name=None, parent=None):
         else:
             self.root = self.parent.root
             self.name = os.path.join(self.parent.name, name)
-        # Initialize data
-        if isinstance(data, dict):
-            self.update(data)
-        else:
+        # Update data from directory path (string) or data dictionary
+        if isinstance(data, type("")):
```
jscotka, line 110: Why you use `type("")` and not `str` directly, is there any  rule, why this is better?

jscotka (approved):
Just one nit pick question, but it seems very well.

I've expected less changes for this :-)

fmf/base.py
psss: That's because of python 2/3 compatibility.

psss:
> I've expected less changes for this :-)

Yes, me too. And much less time as well ;-)

fmf/base.py
jscotka: Ah, thanks for explanation. I though it is some reason like this.

lukaszachy (approved):

psss:
Long live tests! The failure on `epel-7` revealed a Python 2 problem: The tree path could be both `str` and `unicode` which the `isinstance(data, type(""))` did not cover. Adjusted the condition to explicitly check for `None` instead.

psss:
The fresh `epel-7` and `rawhide` failures are expected. Ready do merge.

lukaszachy (approved):

PR #136, opened 2021-04-27
plugin metadata loader with pytest and bash example

jscotka:
**It is WIP, so please no stylish comments, I expect more fundamental comments of this architecture design**

example of plufing for FMF to discover pytest tests and bash.

### cool features:
  - able to add additional metadata directly via python decorators
  - mixing of FMF and plugin metadata together
  - transformation of pytest marks to FMF metadata
    - `mark.skip` transformed to enabled: False
    - custom marks are transformed to tags 
  -  Added `.fmf/config` sa way how to configure plugins
  
output after `cd tests/tests_plugin/`:
```
$ PLUGINS=../../fmf/plugins/pytest.py,../../fmf/plugins/bash.py  fmf show

/pure_fmf
author: Jan Scotka <jscotka@redhat.com>
summary: Pure FMF test case
test: ./runtest.sh

/runtest
author: Jan Scotka <jscotka@redhat.com>
summary:  Basic test case: 1
tag:  tier1
test: ./runtest.sh

/test_basic/TestCls/test
author: Jan Scotka <jscotka@redhat.com>
summary: test_basic.py TestCls test
test: python3 -m pytest -m '' -v test_basic.py::TestCls::test

/test_basic/test_fail
author: Jan Scotka <jscotka@redhat.com>
summary: test_basic.py test_fail
test: python3 -m pytest -m '' -v test_basic.py::test_fail

/test_basic/test_parametrize[a]
author: Jan Scotka <jscotka@redhat.com>
summary: test_basic.py test_parametrize[a]
test: python3 -m pytest -m '' -v 'test_basic.py::test_parametrize[a]'

/test_basic/test_parametrize[b]
author: Jan Scotka <jscotka@redhat.com>
summary: test_basic.py test_parametrize[a]
test: python3 -m pytest -m '' -v 'test_basic.py::test_parametrize[b]'

/test_basic/test_parametrize[c]
author: Jan Scotka <jscotka@redhat.com>
summary: test_basic.py test_parametrize[a]
test: python3 -m pytest -m '' -v 'test_basic.py::test_parametrize[c]'

/test_basic/test_pass
author: Jan Scotka <jscotka@redhat.com>
summary: This is basic testcase
tag: Tier1
test: python3 -m pytest -m '' -v test_basic.py::test_pass

/test_basic/test_skip
author: Jan Scotka <jscotka@redhat.com>
enabled: False
summary: test_basic.py test_skip
test: python3 -m pytest -m '' -v test_basic.py::test_skip
```

### TODOs
 - probably extension is not enoung, could be regexp instead of ".py" could be e.g. `".*tests/.*\py"`
 - how to behave if no data found, maybe return None and it will behave as tree node creating is skipped. (`None != {}` in case {} create the Node)
 - what does with modification? not easy to write data back to non fmf files, e.g. pytest tests. e.g. adding `extra-nitrate` key
   - solution could be add another required method for plugin to write data back for each plugin. Could be harder to parse python code and e.g. add decorator to method. if not exists (but some magic with inspect and looking into function definition source code lines and put there another line with decorator may work well and not so hard to implement `inspect.getfile(item.function), inspect.getsourcelines(item.function)[1]` and insert new line into the file if not found inside other decorators already in:  [`line for line in inspect.getsourcelines(item.function)[0] if re.match(r"\s*@FMF",line)]`)
   - Or store owerwritten data to separate FMF file  using merging+scattering feature, could be easy. (But sorting and file priorities could  cause issues here)

### EDIT
 - Implemented how to modify items for plugin (pytest plugin example for new items)
 
### Focus items for review:
 - generic config file inside .fmf dir for fmf (now used just for plugin loading)
 - changed sorting of files order `(main.fmf, ..., somename.*, somename.fmf (fmf as last in case of same names), ... )
 - modification overriding to FMF files when plugin does not support write method
 - loading plugins by file path or via python way via modules like `a.b.c`
 - using env variable to load (override plugins) `PLUGINS` env var (could be renamed to whatever)
 - where to store plugins (some could be part of FMF project) or strictly as separated in project? (then TMT needs to have some possibility to install some pip module before runnig `discover` on target system or part of target repo)
 - what about extension of plugin by another config (originally I have way in my pytest plugin to use own config, to be able to tweak output format, e.g. say that every `TMT.tag` will be translaated to `tag+`) e.g. I'm configuring `test:` item via this in various ways (another for downstream and upstream)


jscotka:
added to hacking session as topic.

fmf/plugin_loader.py
```diff
@@ -0,0 +1,105 @@
+import importlib
+import inspect
+import os
+import re
+from functools import lru_cache
```
lukaszachy, line 5: fmf should support also python2.7 where this is not available.

fmf/plugin_loader.py
```diff
@@ -0,0 +1,105 @@
...
+                    hierarchy,
+                    modified_dict,
+                    append_dict),
+                stream=fd)
+
+
+@lru_cache(maxsize=None)
+def enabled_plugins(*plugins):
+    plugins = os.getenv(PLUGIN_ENV).split(
+        ",") if os.getenv(PLUGIN_ENV) else plugins
```
lukaszachy, line 68: I believe TMT splits on space, let's be consistent.

fmf/constants.py
```diff
@@ -0,0 +1,11 @@
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+#  Constants
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+SUFFIX = ".fmf"
+MAIN = "main" + SUFFIX
+IGNORED_DIRECTORIES = ['/dev', '/proc', '/sys']
+# comma separated list for plugin env var
+PLUGIN_ENV = "PLUGINS"
```
lukaszachy, line 9: `FMF_PLUGINS` to mitigate name conflict

fmf/plugins/pytest/constants.py
```diff
@@ -0,0 +1,13 @@
+CONFIG_POSTPROCESSING_TEST = "test_postprocessing"
+PYTEST_DEFAULT_CONF = {
+    CONFIG_POSTPROCESSING_TEST: {
+        "test": """
+cls_str = ("::" + str(cls.name)) if cls.name else ""
+escaped = shlex.quote(filename + cls_str + "::" + test.name)
+f"python3 -m pytest -m '' -v {escaped}" """
```
lukaszachy, line 7: I'm not sure about pytest as a require :/ As it is now it is a 'hidden' require - you need to remember. But on the other hand it leaves up to user whether rpm or pip install is better.

fmf/plugins/pytest/plugin.py
```diff
@@ -0,0 +1,310 @@
...
+        func = item.function
+        for marker in item.iter_markers():
+            key = marker.name
+            args = marker.args
+            kwargs = marker.kwargs
+
+            if key == "skip":
+                TMT.enabled(False)(func)
+            elif key == "skipif":
+                # add skipif as tag as well (possible to use adjust, but
```
lukaszachy, line 208: Could skipif be evaluated at this point? Imho `mark.skipif(False)` should end up as `enabled: False` and not `tag: SKIP False`

lukaszachy (commented):
Few comments for now, I need to understand the code for the rest :)

Also an exception when PLUGINS contains broken path could be better.

fmf/plugins/pytest/plugin.py
jscotka: It is not possible, because evaluation will be done in sense of TMT on `guest`, although TMT discovery is done on `host` system.
This `skipIf` is little bit problem to incorporate somehow easily into FMF metadata, but this tag was the simplest solution and probably the easy to understand.
Probably the beast one could be to trasfer it to something like:
```
adjust:
    when:
      interpreter: python
      code: False
    enabled: False
```
and execute adjust condition on guest system. But nothing similar is supported right now.

fmf/plugins/pytest/constants.py
jscotka: yes, it is now hidden, you are rigt, and theoretically it is also why I've let it configurable, not hardcoded.  in my project `fmf_metadata` I have own configs for upstream and dowstream data generation.  and in case you have just unittests then you an execute it directly via python `unittest` executor, or via some `nose` so that pytest is just default way.

fmf/constants.py
jscotka: Yes, sounds reasonable.

fmf/plugin_loader.py
jscotka: Yes, I can change it to whatever more consistent solution, I also thought to use `:` as std path separator, or ` ` but ` ` is worse thanks to it is more common to have spaces in paths, than `:` or `,` chars

fmf/plugins/pytest/tmt_semantic.py
```diff
@@ -0,0 +1,273 @@
...
+                post_mark=post_mark,
+                )
+
+        return inner
+
+    def __getattr__(cls, name):
+        return cls._set_fn(name)
+
+
+class TMT(metaclass=__TMTMeta):
```
happz, line 76: One thing I can say I disagree with right away: this approach taints fmf with a custom code to make a single library user happy and makes fmf do things that are IMO well beyond its scope.

First, it makes adding and changing tmt specifications harder, adding a new place that needs to be updated when specification changes. But, more importantly: I don't think it's up to fmf to inspect the input or validate it in any way.

Is fmf validating or converting tmt keys when reading them from a genuine fmf file? I don't think that's the case, and it should not happen here either. If I put `tag: false` into my `main.fmf` of a test, fmf library will not complain, it will not touch the value, it will not check against a tmt specification schema, it will read it, put it into a fmf tree, and it will deliver the key and its value to tmt. End of story. It's tmt who's responsible for validating the input and reporting its type is invalid and unsupported.

Would it be possible to make the pytest plugin as "dumb" as the actual fmf processing, i.e. treating the input values as opaque blobs of key & value bits, and stacking them into a final fmf tree? I bet it would be possible with some clever Python magic, either have an open decorator accepting all kinds of kwargs or some dynamic decorators - but I **strongly** believe the openness here is a **good thing**, not a bug - fmf should not care about key names or value types, it does not care about them when reading them from `*.fmf` files.

```python
# Probably the simples, but less readable:
@fmf(
    tag=['foo', 'bar'],
    adjust=[
        {...}
...)

# Or maybe allow YAML-in-a-string and parse it as if read from an fmf file?
@fmf(
"""
author: foo@bar.baz
tag:
  - foo
  - bar
adjust:
...
""")

# Or with some __getattr__ magic, this could be possible by creating dynamic decorators aware of their name, i.e. the key:
@fmf.author('foo@bar.baz')
@fmf.tag(['foo', 'bar'])
...
```

tmt is and should remain the sole owner of the semantics. Just like any other tool using fmf to read data. It's not up to fmf to judge, nobody asked fmf's opinion on keys from `*.fmf` files, why would it matter here?

Should a competing test case management tool appear, build on top of fmf too, but with a different set of keys, would fmf learn its semantics as well? And when another two appear, these brand new and cool `mmt` and `ttm` tools, will fmf again learn about their semantics? I don't think that's a good road to take :)

jscotka: you are right, semantics, could be part of another tool. I've Implemented it as one module, because in this case it is all in one solution. I've wanted to show complete solution. My idea was that in FMF there will be just support for plugin, and implementation of plugins will be separate project or part of TMT. But plugin solution alone is hard to test itself. so that complet solution also proves how it works and it also show usecase in TMT or whatever else. I'm familiar to split it. But we have to agreed on the solution for plugin. and improve this part to have it testable as much as possible and when we found it is good enough then I'll split it.

jscotka:
will be replaced by partial changes: 
phase 1: https://github.com/teemtee/fmf/pull/303

PR #144, opened 2021-11-02
Drop support for Python 2

psss:
Remove related stuff from the spec file.
Remove Python 2 only code, disable epel-7 packit.
Remove irrelevant adjust rule from unit tests.
Get rid of other `unicode` related leftovers.
Use default encoding for running commands.
Explicitly mention the `utf-8` encoding in the docs.
Remove unnecessary `object` from class definitions.

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -18,7 +18,6 @@ jobs:
       targets:
       - fedora-all
       - epel-8
-      - epel-7
 
   - job: copr_build
     trigger: commit
@@ -27,7 +26,6 @@ jobs:
       targets:
       - fedora-all
       - epel-8
-      - epel-7
       list_on_homepage: True
       preserve_project: True
       owner: psss
--- a/docs/concept.rst
+++ b/docs/concept.rst
@@ -70,7 +70,8 @@ files on the filesystem:
 * main.fmf
 
 Special file name ``main.fmf`` works similarly as ``index.html``.
-It can be used to define the top level data for the directory.
+It can be used to define the top level data for the directory. All
+metadata files are expected to be using the ``utf-8`` encoding.
 
 
 Attributes
--- a/fmf.spec
+++ b/fmf.spec
@@ -9,30 +9,8 @@ BuildArch: noarch
 URL: https://github.com/psss/fmf
 Source0: https://github.com/psss/fmf/releases/download/%{version}/fmf-%{version}.tar.gz
 
-# Depending on the distro, we set some defaults.
-# Note that the bcond macros are named for the CLI option they create.
-# "%%bcond_without" means "ENABLE by default and create a --without option"
-
-# Fedora or RHEL 8+
-%if 0%{?fedora} || 0%{?rhel} > 7
-%bcond_with python2
-%bcond_with oldreqs
-%bcond_with englocale
-# RHEL 7
-%else
-%bcond_without python2
-# The automatic runtime dependency generator doesn't exist yet
-%bcond_without oldreqs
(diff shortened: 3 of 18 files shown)


psss:
@lachmanfrantisek, I've removed `epel-7` from the `.packit.yaml` but the `epel-7` build seems to be included in the latest packit [job](https://copr.fedorainfracloud.org/coprs/packit/psss-fmf-144/build/2917945/). Is that expected?

lachmanfrantisek:
> @lachmanfrantisek, I've removed epel-7 from the .packit.yaml but the epel-7 build seems to be included in the latest packit job. Is that expected?

I thought we had already fixed this. (@mfocko I recall we've discussed this).

@psss It should affect only this PR and will not be used once merged...

mfocko:
@psss It's twice there, once for `pull_request` trigger and second time for builds _on push to `master`_ (← this one I see removed in diff for PR).

We've also changed the behavior, such that we remove targets only on projects **created and maintained** by Packit user, so if you want to remove `epel-7` from the Copr repo you own (where you build everything pushed to `master`), you need to do so yourself. However it shouldn't build for it, cause it checks only if the targets from config are enabled and builds those. I won't mind if you check it ;)

psss:
@lachmanfrantisek, @mfocko, thanks for quick review and hints. I've completely forgotten that the target config is there twice. After removing the second one everything seems to be working as expected. Sorry for the noise.

lukaszachy (changes requested):
I think we can remove also `adjust` part of `tests/unit/main.fmf` as there is no way python2 tests will pass

jscotka (commented):
fixes seems more well,  but not sure if there are some missing fixes inside rest of the code. e.g.some `unicode(item)` notations inside utils e.g. on line  https://github.com/psss/fm/blob/master/fmf/utils.py#L316 what seems related to Python 2 I think. But maybe it is not hamffull and should not cause anything important.

lukaszachy:
Good point @jscotka , `unicode(item)` raises `NameError:` for python3 so except cause will be hit each time

psss:
@lukaszachy, @jscotka, good points! All should be covered now.

jscotka (approved):
Now, I'm fine with changes. after removing old `unicode()` stuff

psss:
> I think that also "class Tree(object)" should not be used inside python3 just "class Tree:"

@jscotka could you please elaborate a bit more why `object` should not be used?

jscotka:
> > I think that also "class Tree(object)" should not be used inside python3 just "class Tree:"
> 
> @jscotka could you please elaborate a bit more why `object` should not be used?

I had that feeling, maybe there were some hints or docs what I've read that somewhere, now when looking inside: https://stackoverflow.com/questions/4015417/why-do-python-classes-inherit-object it seems that there were difference inside Python2, and in python3 there syntax `class X:` and `class X(object):` is totally equivalent, so that reason why to not use it is that you saves 8 chars.

psss:
@jscotka, thanks for clarification. Unnecessary `object` strings removed from class definitions. @lukaszachy anything else to change or are we ready for squash and merge?

lukaszachy (approved):

lukaszachy:
LGTM.

jscotka (approved):
LGTM

FrNecas (approved):
The code's a lot cleaner now, I like it!

psss:
@lukaszachy, @jscotka, @FrNecas, thanks much for the review! Rebased, squashed, now waiting for tests.

PR #145, opened 2021-11-03
Migrate yaml parsing to ruamel.yaml, remove hacks

psss:
Migrate yaml parsing code to `ruamel.yaml`.
Update spec file and `setup.py` accordingly.
Remove a bunch of old hacks, adjust tests.

### DIFF
--- a/fmf/utils.py
+++ b/fmf/utils.py
@@ -12,8 +12,9 @@ import warnings
 from io import StringIO
 from pprint import pformat as pretty
 
-import yaml
 from filelock import FileLock, Timeout
+from ruamel.yaml import YAML, scalarstring
+from ruamel.yaml.comments import CommentedMap
 
 import fmf.base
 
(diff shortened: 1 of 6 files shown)


fmf/utils.py
```diff
@@ -808,51 +783,27 @@ def run(command, cwd=None, check_exit_code=True, env=None):
...
+    # Set formatting options
+    yaml = YAML()
+    yaml.indent(mapping=4, sequence=4, offset=2)
+    yaml.default_flow_style = False
+    yaml.allow_unicode = True
+    yaml.encoding = 'utf-8'
+    yaml.width = width
+
+    # Convert multiline strings
+    scalarstring.walk_tree(data)
```
psss, line 799: @FrNecas, I've copied this from your `tmt` migration to `ruamel` but I'm not sure: Why exactly is this needed? Does it modify `data`? Should we create a deep copy to prevent any changes to `data`? I tried to apply it on a `dict` which contained a multiline string but could not see any change...

FrNecas: @psss I guess I should have been more explicit in the comment. Yes indeed, it should modify the `data` as far as I remember. There was a hack in tmt that took care of this for PyYAML (so that multiline strings would be formatted accordingly) so I wanted to keep the behaviour the same and found this: https://stackoverflow.com/a/50853272 .

According to the comment and docs, the `walk_tree` method should convert string to `LiteralScalarString` if the element is a multiline string which is what we wanted due to formatting. I don't know what you've tried but a quick test shows this:

```
>>> from ruamel.yaml import scalarstring
>>> data = {'a': 'a\nb\nc\n'}
>>> type(data['a'])
<class 'str'>
>>> scalarstring.walk_tree(data)
>>> type(data['a'])
<class 'ruamel.yaml.scalarstring.LiteralScalarString'>
>>> data['a']
'a\nb\nc\n'
```
The `LiteralScalarString` class probably has a set `__str__` formatting but as you can see, it was transformed. As for your last question, I don't really know. It may make sense to create a copy in fmf so that the formatting is separate from the inner logic? On the other hand, it inherits from `str` class so it behaves like a normal string just has some more logic related to ruamel so it may not be a problem at all to do this without copying. I will take a closer look tomorrow or on Friday.

FrNecas: On the other hand, users of fmf as a library may expect that it keeps the formatting the way it was? It's definitely up for a discussion :)

psss:
All tests passed except `/tests/test/lint` from the integration plan which is the expected change of behaviour we discussed on the hacking session:

```
cd git/tmt/tests/test/lint/data/
tmt test lint old-yaml
```

Before it was:

```
/old-yaml
pass test script must be defined
pass directory path must be absolute
pass directory path must exist
warn summary is very useful for quick inspection
warn seems to use YAML 1.1 syntax which will be deprecated in tmt 2.0
pass correct attributes are used
```

And now we get error during individual attribute check for correct type:

```
Invalid 'enabled' in test '/old-yaml' (should be a 'bool', got a 'str').
```

I guess this is acceptable. The error is clear and clearly points to the source of the problem. Thoughts?

fmf/utils.py
```diff
@@ -782,37 +783,28 @@ def run(command, cwd=None, check_exit_code=True, env=None):
...
+    yaml = YAML()
+    yaml.indent(mapping=4, sequence=4, offset=2)
+    yaml.default_flow_style = False
+    yaml.allow_unicode = True
+    yaml.encoding = 'utf-8'
+    yaml.width = width
 
-yaml.add_representer(str, repr_str, Dumper=yaml.SafeDumper)
+    # Make sure that multiline strings keep the formatting
+    data = copy.deepcopy(data)
```
jscotka, line 799: this line here is to avoid `YAML` referencing abilitiers?

jscotka (approved):
Fix seems very clean, LGTM.
Just some advices: 
   * one question comment inside  the code
   * Would like to see also test yaml file with comments, to see if orderings works as expected also for comments.

fmf/utils.py
psss: The `walk_tree()` method used on the next line modifies all strings which have multiple lines. Although this change to `LiteralScalarString` should not affect the string-related methods and should be relatively ok, I think it is safer not to modify provided `data` at all.

jscotka: okay, thanks for explanation

psss:
Thanks for the reviews!

PR #147, opened 2021-11-11
Version as fmf flag and module attribute

lukaszachy:
"Got inspired" from tmt.


psss (approved):
Looks good, except that the `version` was not actually printed. Should be fixed by 05c79ec.

PR #148, opened 2021-11-23
Enable centos-stream-9 for copr builds and testing

psss:
Let's see what happens ;-)

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -18,6 +18,7 @@ jobs:
       targets:
       - fedora-all
       - epel-8
+      - epel-9
 
   - job: copr_build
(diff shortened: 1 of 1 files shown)


psss:
Fails on a missing dependency:

    No matching package to install: 'python3-filelock'

psss:
Installation on `centos-stream-9` fails because of a missing dependency:

      - nothing provides python3.9dist(filelock) needed by python3-fmf-1.0.0-1.20220328152012419017.pr148.6.g1cbc082.el9.noarch

@thrix, it seems that the `epel-9` repository is not enabled on the guest. Could you please have a look?

PR #151, opened 2022-02-04
Fixes blocking with symbolic link loops.

odkq:
When uncompressing a data file that contained a copy of the /sys/ tree of a SUT
in a subdirectory of the working tmt dir, I noticed that symbolic link loops
are not detected while traversing the directories discovering .fmf files

Easily reproducible by doing in an empty directory:

    tmt init .
    ln -s /sys/ .
    tmt     # blocks

Debugging with pdb:

    $ python3 -m pdb ~/src/github.com/odkq/tmt/bin/tmt
    > ~/src/github.com/odkq/tmt/bin/tmt(3)<module>()
    -> import click
    (Pdb) continue
    ^C
    Program interrupted. (Use 'cont' to resume).
    > /usr/lib64/python3.10/os.py(377)_walk()
    -> except OSError:
    (Pdb) up
    > ~/.local/lib/python3.10/site-packages/fmf/base.py(411)grow()
    -> dirpath, dirnames, filenames = next(os.walk(path))
    (Pdb) print(path)
    /tmp/test-tmt/sys/block/dm-0/bdi/subsystem/0:66/[... many more]/subsystem/8:0/subsystem/253:0/subsystem/253:0/subsystem/btrfs-1/subsystem/0:66/subsystem

Added a list that is copied from parent Trees to children that keeps track of
the destination path of the symbolic links visited and do not walk again once
visited to avoid the above situation.

### DIFF
--- a/fmf/base.py
+++ b/fmf/base.py
@@ -58,6 +58,13 @@ class Tree:
         # (needed to prevent removing nodes with an empty dict).
         self._updated = False
 
+        # Store symlinks in while walking tree in grow() to detect
+        # symlink loops
+        if parent is None:
+            self._symlinkdirs = []
+        else:
+            self._symlinkdirs = parent._symlinkdirs
+
         # Special handling for top parent
         if self.parent is None:
             self.name = "/"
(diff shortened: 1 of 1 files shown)


lukaszachy (approved):
With /sys symlink it is super easy to reproduce. However I can't replicate using "normal" link structure. Either way PR looks good and I'll merge once tests are green (looks as infra issues). 
Thank you for the PR.

Issue #154, opened 2022-03-04
[RFE] Allow allways True condition for adjust: when


kkaarreell:
It should be possible to have a condition that is always True, e.g.
```
adjust:
  when: True
```
or maybe even not having `when` mandatory (asssumed `when: True`) when missing.

This would be useful in case there are multiple `when` conditions in adjust and some should be applied only sometimes while some other always.

psss:
Just to clarify the use case: Is this just for debugging purposes? Or what would be motivation for always adjusting metadata instead of just setting the value directly?

Regarding the naming, `when: true` sounds fine, `when: always` could also be a nice explicit option. Agree that missing `when` should mean adjust always.

lukaszachy:
If this is for debugging purposes than I think `when: false` might be useful as well (as opposed to commenting/deleting whole rule)

kkaarreell:
Not for debugging purposes. Simply I wanted to build test list from "blocks' while some blocks would be conditioned.
It is not because the test won't be able to run in some environment, it is just that it won't provide any benefit.

My original use case was like this:
```
/e2e:

  summary: run keylime e2e tests

  prepare:
    how: shell
    script: bash -c 'ln -s $(pwd) /var/tmp/keylime_sources'

  discover:
    how: fmf
    url: https://github.com/RedHat-SP-Security/keylime-tests
    ref: main
    test: 
     - /setup/configure_tpm_emulator
     - /setup/install_upstream_keylime

  adjust:
   # prepare step adjustments
   - when: distro == centos-stream-9
     prepare+:
       script: yum -y install https://dl.fedoraproject.org/pub/epel/epel-release-latest-9.noarch.rpm

   - when: distro == centos-stream-8
     prepare+:
       script: yum -y install https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.noarch.rpm

   # discover step adjustments

   # enable code coverage measurement on F35
   - when: distro == fedora-35
     discover+:
       test+:
        - /setup/enable_keylime_coverage

   # now add all the required tests
   # user workaround for when: always
   - when: sense_of_life is not defined
     discover+:
       test+:
        - /functional/basic-attestation-on-localhost
        - /functional/keylime_tenant-commands-on-localhost
        - /functional/tpm_policy-sanity-on-localhost
        - /functional/db-postgresql-sanity-on-localhost
        - /functional/db-mariadb-sanity-on-localhost
        - /functional/db-mysql-sanity-on-localhost

[...]

PR #156, opened 2022-03-08
Make when in adjust optional, add true|false expr.

lukaszachy:
One can ommit when: key to have always passing adjust rule.
Rule can contain 'true' and 'false' as expression.

Solves #154

### DIFF
--- a/docs/context.rst
+++ b/docs/context.rst
@@ -68,7 +68,7 @@ Let's demonstrate the syntax on a couple of real-life examples::
     # disable adjust rule (e.g. during debugging / experimenting)
     false and <original rule>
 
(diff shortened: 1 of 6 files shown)


psss (approved):
Looks good. Added just a few minor adjustments in 307e5e4. @lukaszachy are the changes ok?

lukaszachy:
> Looks good. Added just a few minor adjustments in [307e5e4](https://github.com/teemtee/fmf/commit/307e5e4d809e322465d6f13e3cedf79ed9fd15c2). @lukaszachy are the changes ok?

LGTM, Thanks for the adjustements.

PR #158, opened 2022-03-23
Add support for prepending items during merging

### LINKED ISSUE #155: Allow prepending in adjust
It would be nice to be able to add something at the beginning, not an end.

I do not have exact syntax in mind but something like this is working for me.
```
--- /usr/lib/python3.6/site-packages/fmf/base.py.orig	2021-11-11 11:37:06.000000000 +0100
+++ /usr/lib/python3.6/site-packages/fmf/base.py	2022-03-04 11:23:42.689937201 +0100
@@ -139,7 +139,7 @@
         except ValueError:
             raise utils.FormatError("Invalid version format")
 
-    def _merge_plus(self, data, key, value):
+    def _merge_plus(self, data, key, value, prepend=False):
         """ Handle extending attributes using the '+' suffix """
         # Nothing to do if key not in parent
         if key not in data:
@@ -151,7 +151,10 @@
             return
         # Attempt to apply the plus operator
         try:
-            data[key] = data[key] + value
+            if prepend:
+                data[key] = value + data[key]
+            else:
+                data[key] = data[key] + value
         except TypeError as error:
             raise utils.MergeError(
                 "MergeError: Key '{0}' in {1} ({2}).".format(
@@ -189,6 +192,8 @@
             # Handle special attribute merging
             if key.endswith('+'):
                 self._merge_plus(data, key.rstrip('+'), value)
+            elif key.endswith('+<'):
+                self._merge_plus(data, key.rstrip('+<'), value, prepend=True)
             elif key.endswith('-'):
                 self._merge_minus(data, key.rstrip('-'), value)
             # Otherwise just update the value
```

psss:
Implement the special `+<` suffix which allows to prepend values at the
beginning of the list when merging values.

Resolves #155.

### DIFF
--- a/docs/features.rst
+++ b/docs/features.rst
@@ -89,6 +89,18 @@ Exception ``MergeError`` is raised if types are different. When
 the ``+`` suffix is applied on dictionaries ``update()`` method is
 used to merge content of given dictionary instead of replacing it.
 
+The special suffix ``+<`` can be used to prepend values instead of
+appending them. This might be handy when adjusting lists::
+
+    steps:
+      - one
+      - two
+      - three
+
+    /complete:
+        steps+<:
+          - zero
+
 In a similar way, appending a ``-`` sign will reduce or remove
 parent value from parent's attribute (which has to be defined)::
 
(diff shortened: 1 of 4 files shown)


psss:
Not completely sure about the `+<` suffix but cannot think of anything better. @lukaszachy, @FrNecas, any ideas?

PR #159, opened 2022-04-05
special no inheritance mark

jscotka:
### Cretead improvement, to have some items without inheritance in tree
 - uses special characted **!** at the end of key e.g. `key!: some value`

### There are two possible solution: 
already implemented in two inclueded commits
 - let special character at the end, so that key is `key!` in data.
   - PROS:
     - easy to find the item
     - easy to search
     - easy to lint
   - CONS:
     - special character may be confusing comparing to removing `+` and `-` elements from final tree
     - may contain `key` and `key!` together - may confuse users
 -  remove special character and let it as normal key
    - PROS
      - easier for user to understand it, it is special  char, and will be removed
      - automatically it solve also removing keys as side effect, so it stop inheritance, if there is defined `X` in parent and in some child `X!` then next child will have `X` undefined
    - CONS
      - not so clean from perspective that you may mix `X` and `X!` so that you will use these inherited and uninherited items together and not easy to search for special key with `X!` to know that it is defined on specifix place (may be workarounded but not nice)
      - When user remove `!` it will behave in same way, so that tool has to know, that in node there is special attribute what contains, these unique keys for node.


lukaszachy:
@psss @jscotka We have implemented `inherit: false` . Should this be dropped?

jscotka:
> @psss @jscotka We have implemented `inherit: false` . Should this be dropped?

yes, we can drop it. it is similar.

PR #165, opened 2022-05-31
Workaround s390x ruamel read issue

### LINKED ISSUE #164: Fail to parse fmf files on s390x
The following error is raised when trying to parse an fmf file:

```
unacceptable character #x0000: control characters are not allowed
```

Example:

```
# git clone https://github.com/cockpit-project/cockpit-podman.git
# cd cockpit-podman/test/browser/
# fmf show
 ERROR  Failed to parse '/root/cockpit-podman/plans/all.fmf'.
unacceptable character #x0000: control characters are not allowed
  in "/root/cockpit-podman/plans/all.fmf", position 72
```

This problem was introduced on version `1.0.0`

Note:

replacing `YAML(typ="safe").load(datafile)` by `YAML().load(datafile)` seems to solve the problem.

lukaszachy:
Reading file via python and then using yaml parser works
on that architecture

Fixes #164

### DIFF
--- a/fmf/base.py
+++ b/fmf/base.py
@@ -435,7 +435,9 @@ class Tree:
             log.info("Checking file {0}".format(fullpath))
             try:
                 with open(fullpath, encoding='utf-8') as datafile:
-                    data = YAML(typ="safe").load(datafile)
+                    # Workadound ruamel s390x read issue - fmf/issues/164
+                    content = datafile.read()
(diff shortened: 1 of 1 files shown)


lukaszachy:
> what an interesting bug

Indeed but I'm not going to dive into compiled code.

psss (approved):
Thanks for the quick workaround, looks good.

PR #167, opened 2022-06-06
Explicitly convert context value to string

thrix:
Actually writing an integer value in context value
should be supported, `fmf` should take care of the conversion.

Seems this place is the best for that.

See for details https://github.com/teemtee/fmf/issues/168

Signed-off-by: Miroslav Vadkerti <mvadkert@redhat.com>

### DIFF
--- a/docs/context.rst
+++ b/docs/context.rst
@@ -112,8 +112,9 @@ Each Dimension is a view on the Context in which metadata can be
 adjusted. For example it can be arch, distro, component, product
 or pipeline in which we run tests and so on.
 
-Each value is treated as if it was a component with version. Name
-of the dimension doesn't matter, all are treated equally.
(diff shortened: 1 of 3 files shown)


thrix:
@thrix please add some tests coverage :)

psss (changes requested):
Looks good, let's just update the [Dimensions](https://fmf.readthedocs.io/en/latest/context.html#dimensions) section with a mention that dimension values are always converted to `str`.

thrix:
> Looks good, let's just update the [Dimensions](https://fmf.readthedocs.io/en/latest/context.html#dimensions) section with a mention that dimension values are always converted to `str`.

Should be done @psss

psss (approved):
Looks good, squashed.

PR #170, opened 2022-06-08
Implement a directive for disabling inheritance

### LINKED ISSUE #14: Define a way how to undefine an attribute
It would be useful to be able to undefine an inherited attribute. For example having an extra directory with helper files which should not be considered as tests. Perhaps something like this?

    test: None

psss:
As the first step for the special `fmf` directives let's allow to
completely disable inheriting data from parent which is currently
the most desired use case.

Fix #14.


thrix (approved):
LGTM

lukaszachy (approved):

happz (approved):
Filed https://github.com/teemtee/fmf/issues/171 to address the reinvention of the validation process. Other than this, LGTM.

PR #173, opened 2022-06-08
Use `tree wget -F --dirsfirst` to clarify directory structure

abitrolly:
```
cd tmp
mkdir wget
cd wget
mkdir download protocols recursion smoke
cd protocols
mkdir ftp https http
cd ..
touch main.fmf
cd ..
tree wget -F --dirsfirst
```

### DIFF
--- a/docs/features.rst
+++ b/docs/features.rst
@@ -6,15 +6,17 @@
 Let's demonstrate the features on a simple wget example with the
 following directory structure::
 
-    wget
-    ├── download
-    ├── protocols
-    │   ├── ftp
-    │   ├── http
-    │   └── https
-    ├── recursion
-    └── smoke
+    wget/
+    ├── download/
+    ├── protocols/
+    │   ├── ftp/
+    │   ├── http/
+    │   └── https/
+    ├── recursion/
+    ├── smoke/
+    └── main.fmf
 
+    7 directories, 1 file
 
 Simple
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~


abitrolly:
I am not sure where `runtest.sh` should be. The doc doesn't mention it.

1. Is it one runner for all dirs located in the root?
2. Is it a runner that is placed in each dir?

lukaszachy:
> I am not sure where `runtest.sh` should be. The doc doesn't mention it.
> 
>     1. Is it one runner for all dirs located in the root?
> 
>     2. Is it a runner that is placed in each dir?

Ouch, example was oversimplified. That usage doesn't make much sense without tmt metadata specification. I guess we should come up with something generic enough and not dependent on tmt to stop always using fmf with tmt which makes it difficult to realise what is what.

If I take tmt into account (meaning that `runtest.sh` is command which should be executed for test and as such it should be part of "test" attribute):
- Example is wrong - runtest.sh would need to be executable found on PATH. I guess the current intention was to have `test: ./runtest.sh` instead.
- Each directory should have it's own 'runtest.sh' file
Or you can have single runtest.sh but either `path` or `test` should be adjusted so process executing test has it on expected path (process changes dir according to 'path' attribute and executes whatever is written in 'test' attribute)

abitrolly:
> Ouch, example was oversimplified. 

Oversimplified is better than overcomplicated. :D

> I guess we should come up with something generic enough and not dependent on tmt to stop always using fmf with tmt which makes it difficult to realise what is what.

Something artificial that is not being used in production won't make it better. Tests are at least common scenario both for coders and DevOps teams.

> * Each directory should have it's own 'runtest.sh' file
>       Or you can have single runtest.sh but either `path` or `test` should be adjusted so process executing test has it on expected path (process changes dir according to 'path' attribute and executes whatever is written in 'test' attribute)

By `path` attribute, do you mean `/download:` part here?
```
/download:
    description: Check basic download options
    tester: Petr Šplíchal <psplicha@redhat.com>
    tags: [Tier2, TierSecurity]
    test: ./runtest.sh
    time: 3 min
```
So to clarify, for the `wget/main.fmf` example above, that script that will be executed for this test is `wget/download/runtest.sh`, and `wget/runtest.sh` will be ignored. Is that right?

And if the contents of the `wget/main.fmf` is changed to the below text, the `wget/runtest.sh` will be ignored too, and instead two separate scripts will be run. `wget/download/runtest.sh` and `wget/recursion/runtest.sh`
```
tester: Petr Šplíchal <psplicha@redhat.com>
tags: [Tier2, TierSecurity]
test: ./runtest.sh

/download:
    description: Check basic download options
    time: 3 min
/recursion:
    description: Check recursive download options
    time: 20 min
```

So the question now is how should I rewrite this `wget/main.fmf` to reuse `wget/runtest.sh` in both cases?

abitrolly:
You should have write access to this branch. It should be the best way to put the content right.

lukaszachy:
@psss, @abitrolly What if we linked [tmt tests](https://github.com/teemtee/tmt/tree/main/tests)?
Top [main.fmf](https://github.com/teemtee/tmt/blob/main/tests/main.fmf) defines common attributes, [prepare](https://github.com/teemtee/tmt/blob/main/tests/prepare/main.fmf) changes them for tests for prepare step and [final test](https://github.com/teemtee/tmt/blob/main/tests/prepare/shell/main.fmf) adds just link to coverage and summary.

lukaszachy:
Another idea - would showing final metadata help more to understand what is happening? 
E.g. show that without inheritance one would need to spell out whole:
```
/download:
    description: Check basic download options
    time: 3 min
    tester: Petr Šplíchal <psplicha@redhat.com>
    tags: [Tier2, TierSecurity]
    test: ./runtest.sh
/recursion:
    description: Check recursive download options
    time: 20 min
    tester: Petr Šplíchal <psplicha@redhat.com>
    tags: [Tier2, TierSecurity]
    test: ./runtest.sh
```

psss:
> E.g. show that without inheritance one would need to spell out whole:

This is covered under the [hierarchy](https://fmf.readthedocs.io/en/latest/features.html#hierarchy) section. So we have the comparision.

> For example "/tests/prepare/shell":

The tangible real-life example sounds like a great idea!

abitrolly:
> @psss, @abitrolly What if we linked [tmt tests](https://github.com/teemtee/tmt/tree/main/tests)?

That would create recursive link, which will only worsen the mess in my head.

PR #175, opened 2022-07-11
Select by the source of object

lukaszachy:
E.g. Select all objects which use data defined in a particular file


fmf/base.py
```diff
@@ -558,6 +562,9 @@ def prune(self, whole=False, keys=None, names=None, filters=None,
             if names and not any(
                     [re.search(name, node.name) for name in names]):
                 continue
+            # Select nodes defined by any of the source files
+            if sources and not bool(sources.intersection(node.sources)):
```
psss, line 566:
```suggestion
            if sources and not sources.intersection(node.sources):
```

Or is there a specific reason why explicitly convert to `bool` here?

psss (approved):
Thanks for writing this! Looks very good. Added some trivial adjustments and renamed duplicate test method (going to enable `flake8` to catch similar typos sooner).

fmf/base.py
lukaszachy: Frankly I have not idea.

psss: Ok, removed then.

Issue #169, opened 2022-06-07
Use `--depth 1` when cloning git repos with no `ref` specified


psss:
In order to speed up testing and minimize data transfers let's use `git clone --depth 1` when cloning git repositories and no specific `ref` is requested by the user.

PR #177, opened 2022-07-15
Shallow git clone if no reference is used

lukaszachy:
If later `fetch_repo()` specifies ref,
git fetch --unshallow is called to obtain whole history.

Fixes: #169


fmf/utils.py
```diff
@@ -735,7 +739,18 @@ def fetch_repo(url, ref=None, destination=None, env=None):
...
             # Checkout branch
-            run(['git', 'checkout', '-f', ref], cwd=destination, env=env)
+            try:
+                run(['git', 'checkout', '-f', ref], cwd=destination, env=env)
+            except subprocess.CalledProcessError as error:
+                # Another way to check for shallow clone is to call
+                # `git rev-parse --is-shallow-repository` but it prints true/false
+                if os.path.isfile(os.path.join(destination, '.git', 'shallow')):
+                    # Fetch whole history and try again
+                    run(['git', 'fetch', '--unshallow'], cwd=destination)
```
psss, line 749:
```suggestion
                    run(['git', 'fetch', '--unshallow', 'origin', '+refs/heads/*:refs/heads/*'], cwd=destination)
```

Seems we have to explicitly ask to fetch remote branches as well otherwise the second checkout fails:

    fmf.utils.fetch_repo("https://github.com/teemtee/tmt/", destination="/tmp/fetched")
    fmf.utils.fetch_repo("https://github.com/teemtee/tmt/", destination="/tmp/fetched", ref="fedora")
    error: pathspec 'fedora' did not match any file(s) known to git

Perhaps add a simple test to cover this?

psss (changes requested):
Looks good except that it seems that we have to explicitly ask to fetch remote branches as well.

psss:
> @psss it was quite easy so maybe worth including into upcoming release?

Agreed, let's include it in `1.2`.

fmf/utils.py
lukaszachy: I definitely will add tests (also to check that shallow clone happened). Thanks for suggestion - I've tried with commit hash and that works without suggested change.

lukaszachy: It was not enough, I had to follow https://www.delenamalan.co.za/2019/fetch-all-branches.html

psss (approved):
Thanks for the quick fix! I just fixed typo in comment and going to merge.

PR #180, opened 2022-07-20
Retry if shallow clone fails

lukaszachy:
Some transports doesn't allow shallow clone so retry without depth=1 if
clone failed.

(e.g. cgit for http://server/cgit/repo)

### DIFF
--- a/fmf/utils.py
+++ b/fmf/utils.py
@@ -726,7 +726,15 @@ def fetch_repo(url, ref=None, destination=None, env=None):
             else:
                 depth = []
             if not os.path.isdir(os.path.join(destination, '.git')):
-                run(['git', 'clone'] + depth + [url, destination], cwd=cache, env=env)
+                try:
+                    run(['git', 'clone'] + depth + [url, destination], cwd=cache, env=env)
+                except subprocess.CalledProcessError as error:
+                    # Can't check error message because it is localized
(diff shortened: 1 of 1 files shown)


lukaszachy:
There was 1 tcms case using it (/me even didn't know that it is possible to clone from cgit url). 
I've unpublish fedora update and will do 1.2.1 once this PR is merged.

fmf/utils.py
```diff
@@ -726,7 +726,15 @@ def fetch_repo(url, ref=None, destination=None, env=None):
...
             if not os.path.isdir(os.path.join(destination, '.git')):
-                run(['git', 'clone'] + depth + [url, destination], cwd=cache, env=env)
+                try:
+                    run(['git', 'clone'] + depth + [url, destination], cwd=cache, env=env)
+                except subprocess.CalledProcessError as error:
+                    # Can't check error message because it is localized
+                    if not depth:
+                        # Do not retry if shallow clone was not used
+                        raise
+                    log.debug("Clone failed '{0}', trying without depth".format(error))
```
psss, line 736:
```suggestion
                    log.debug("Clone failed with '{0}', trying without '--depth=1'.".format(error))
```

psss (approved):
Looks good, just proposing a minor wording adjustment.

PR #182, opened 2022-09-06
Fix a typo in the merging minus sign documentation

guoguojenna:
Per my understanding, it might be 'subtracted' instead of 'subtracked'.

Thanks,
Jenna


psss (approved):
Yeah, thanks for fixing the typo. I just updated the commit message to match our [recommendations](https://fmf.readthedocs.io/en/latest/contribute.html#commits) and make the release notes more beautiful ;-)

PR #192, opened 2023-06-12
Add an example for CentOS Stream X distro adjustment

ljavorsk:
There is no mention of CentOS Stream X distro adjustments in the documentation which is now more popular, as RHEL early development is done in those distributions. 
I've added a small example, feel free to expand it however you need.

### DIFF
--- a/docs/context.rst
+++ b/docs/context.rst
@@ -230,3 +230,13 @@ comparison works for some special cases::
     centos-8.4.0 < centos-9 ---> True
     centos-8.4.0 ~< centos-9 ---> True (no minor comparison requested)
     centos-8.4.0 ~< centos-9.2 ---> cannot (minor comparison requested)
+
+If you want to aim for CentOS Stream you can specify the desired version like e.g. ``centos-stream-9``.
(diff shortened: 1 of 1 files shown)


lukaszachy:
@psss Hm, shouldn't we have similar section in tmt  instead? This feels like 'how to write tests' which makes more sense in tmt part. 

Either way thank you for the PR @ljavorsk, just need to figure out where to put it.

ljavorsk:
I've noticed there are already some examples in this doc, that's why I chose it.

psss:
@ljavorsk, thanks for adding this! @lukaszachy is right that the context adjust example fits a bit more to the `tmt` documentation. There are two places where it would make sense to mention `centos-stream-x` values:

* https://tmt.readthedocs.io/en/stable/spec/core.html#adjust
* https://tmt.readthedocs.io/en/stable/spec/context.html#dimension

Thinking about it a bit more, what about rather updating both `fmf` and `tmt` documentation to use `centos-stream` values in the examples instead of `centos-x` as these are (or soon will be) obsoleted?

PR #196, opened 2023-07-25
PoC of profiles support

jscotka:
This PoC shows how the profiles could be solved. 
Main idea there is that there are external ``YAML`` data passed to profile apply. It is similar to adjust and uses similar functions.
external yaml contains definition where it is applied (go thru all leaf nodes) and then ``_merge_special`` method is called and all other keys are applied.
With extension for "``?``" mark to apply something if not defined.

It solves several use cases:

- be able to add/append/remove/override some steps
   - e.g. able to change provision method
   - e.g. add some extra provision steps
- It is generic, so that you have to define the selector ``where`` to define where you want to apply data, e.g. plan is defined as ``execute is defined and discovery is defined``

```yaml
plan1_definition:
  where: execute is defined cc and discover is defined
  # append to prepare phase
  prepare+:
    name: some name
    how: shell
    script: systemctl start httpd
   # create reporting if not defined
  report?:
    how: html
    open: true

   # override provision by own definition
  provision:
    how: virtual
    image: fedora-37


test1_definition:
  where: summary is defined and test is defined
  new_attribute: hello world
  test+: appended
```


LecrisUT:
How is this different then applying the profile yaml as a `main.fmf`? Maybe it has overlap with setting system/user/project metadata file/tree

jscotka:
> at

I've completely forgot this PR from past.
I think my idea was to be able to define or extend something if not defined in project git repo. And avoid to everytime to use command line. 
It is probably something what currently is also TestingFarm doing.
It apply some tool/user level setting how to execute tests, especially define how and where to run tests.

So that it allows to extend some steps, or overwrite it by user/system definition.

PR #203, opened 2023-08-09
Type hints

LecrisUT:
I don't like that the data dict type is an `Any`, but this should work
- [x] `Tree`
- [ ] `Context`
- [ ] `filter


happz:
One quick note: `|` is supported from Python 3.10. I don’t know about other users, but at least tmt would still support Python 3.9. Therefore you might need to use `Union` or wrap unions with `|` with a string (`”int | str”`) to hide them from Python 3.9 interpreter.

LecrisUT:
- Rhel8 still uses `3.6`, which might be a pain to make it compatible
- Rhel9 error is about `|`, so yeah should change that to `Optional` and `Union`
- I am not sure if I got the `Context` typing correct
- Should add mypy to the CI in this PR

happz:
> Should add mypy to the CI in this PR

Absolutely, you can copy the section from tmt's pre-commit configuration, it served there well.

fmf/context.py
```diff
@@ -44,13 +47,13 @@ def __init__(self, origin):
         else:
             self._to_compare = self._split_to_version(origin)
 
-    def __eq__(self, other):
+    def __eq__(self, other: ContextValue):
```
happz, line 50: Linters will complain here: `ContextValue` is being defined right now, it is not yet fully known, therefore it cannot be used as an annotation directly, a forward reference is needed, `other: 'ContextValue'`.

happz: Also, note that the code does handle the possibility of `other` being something completely different than `ContextValue`, I'd say the right type here would be `Any` after all. If the `other` is recognized as `ContextValue` by `isinstance()` below, fine, it's compared, anything else leads to `False`.

LecrisUT: That would be an issue with `from __future__ import annotations`. What python version gives that issue so we can document it?

LecrisUT: > Also, note that the code does handle the possibility of other being something completely different than `ContextValue`

Yeah, I was thinking about that one. But is the usage intent to allow the comparison with non-`ContextValue` types? It would create a bit of ambiguity e.g. is it compared as `ContextValue(other)`?. This way at least the user is encouraged to check the implementation and make a specific type checker independent of that.

fmf/base.py
```diff
@@ -29,10 +33,23 @@
...
 # ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-class Tree:
+class Tree(Mapping[str, Any]):
     """ Metadata Tree """
-
-    def __init__(self, data, name=None, parent=None):
+    parent: Tree | None
+    children: dict[str, Tree]
+    data: dict[str, Any]
```
happz, line 40: This `dict[str, Any]` tends to appear quite often, representing the "raw" node data. I'd suggest adding an alias for it, so we could tell the difference between "`dict` that holds node data" and "`dict` that is just `dict` with `str`/`Any` content". Let's say `TreeData` or something similar? `data: TreeData` would then be easier to track, and it would make method signatures easier to understand: `child(self, name: str, data: dict[str, Any] | str | None, source: str | None = None) -> None:` vs `child(self, name: str, data: TreeData | None, source: str | None = None) -> None:` where it's immediately obvious what this `data` does actually mean.

LecrisUT: Good point. Also can it be further refined to only contain json-like types. Maybe:
```
DataType = int | float | str
ListDataType = list[DataType | ListDataType | DictDataType]
DictDataType = dict[str, DataType | ListDataType | DictDataType]
TreeData = dict[str, DataType | ListDataType | DictDataType]
```

fmf/context.py
happz: > That would be an issue with from __future__ import annotations. What python version gives that issue so we can document it?

Well, IIUIC, every single one of them. When this bit of code is spotted by the Python interpreter, there is no object called `ContextValue` yet, because it's being defined. This is not about Python versions, it's natural, `ContextValue` is not yet complete and therefore it cannot be referenced by code like this. it's exactly the same situation as the following:

```python
class Foo:
    A = 1
    B = Foo.A
```

Python should throw this back at you with "name 'Foo' is not defined" on line 3 (`B = Foo.A`).

>> Also, note that the code does handle the possibility of other being something completely different than ContextValue

> Yeah, I was thinking about that one. But is the usage intent to allow the comparison with non-ContextValue types? It would create a bit of ambiguity e.g. is it compared as ContextValue(other)?. This way at least the user is encouraged to check the implementation and make a specific type checker independent of that.

I don't know what the intent was, I just see the code, and the code clearly accepts situations where `other` can be something else than `ContextValue`. It even makes a special branch for comparing other types with `ContextValue`. Both the annotation and code should be aligned, therefore, as I see it, we should either change the annotation to allow wider inputs or make the non-`ContextValue` case a hard error. If the annotation states "only `ContextValue` allowed", returning `False` for other values makes the error silently disappear.

fmf/base.py
happz: Yep, definitely possible. Although for the beginning, I for one would be happy with just `TreeData = dict[str, Any]`, exactly because it can be tightened later.

fmf/base.py
```diff
@@ -794,3 +822,18 @@ def __getitem__(self, key):
...
+        for c in self.children:
+            yield f"/{c}"
+        for d in self.data:
+            yield d
+
+    def __contains__(self, item: str):
+        if item.startswith("/"):
+            return item[1:] in self.children
+        else:
+            return item in self.data
```
LecrisUT, line 839: These interfaces I am not sure about. Should they be combined here, or just output the `children` data. In particular `len()` feels dubious.

LecrisUT: Also missing interfaces: `keys`, `values`, `items`, `has_key` and probably equivalents to distinguish between data and children

fmf/context.py
LecrisUT: On my side, `python 3.11` + `PyCharm`, there is no error as long as `__future__` is included. I am not sure if that was introduced in `3.10` or `3.9`. Will have to check around on that.

> Both the annotation and code should be aligned, therefore, as I see it, we should either change the annotation to allow wider inputs or make the non-ContextValue case a hard error. If the annotation states "only ContextValue allowed", returning False for other values makes the error silently disappear.

Indeed we should get a clarification on this. It is particularly a problem if you are comparing `neq` and such.

happz: > On my side, python 3.11 + PyCharm, there is no error as long as __future__ is included. I am not sure if that was introduced in 3.10 or 3.9. Will have to check around on that.

Ah, got it, I misunderstood. I believe it should be possible to enable the postponed evaluation of annotations since Python 3.7.

fmf/base.py
```diff
@@ -24,15 +35,40 @@
...
 
+# TypeHints
+RawDataType: TypeAlias = Union[None, int, float, str, bool]
+ListDataType: TypeAlias = list[Union[RawDataType, 'ListDataType', 'DictDataType']]
+DictDataType: TypeAlias = dict[str, Union[RawDataType, ListDataType, 'DictDataType']]
+# Equivalent to:
+# JSON: TypeAlias = dict[str, "JSON"] | list["JSON"] | str | int | float | bool | None
+DataType: TypeAlias = Union[RawDataType, ListDataType, DictDataType]
+TreeData: TypeAlias = dict[str, DataType]
+TreeDataPath: TypeAlias = Union[TreeData, str]  # Either TreeData or path
```
LecrisUT, line 46: I think the naming can be confusing here between `TreeData`, `TreeDataPath` and `DataType`. Any suggestions?

LecrisUT:
MyPy is up and there are issues to resolve

LecrisUT:
All tests are green here right now. Any review please?

fmf/base.py
```diff
@@ -92,10 +133,10 @@ def __init__(self, data, name=None, parent=None):
         if self.parent is None:
             self.inherit()
 
-        log.debug("New tree '{0}' created.".format(self))
+        log.debug(f"New tree '{self}' created.")
```
happz, line 136: `format()` => f-string change seems to be unrelated to adding type annotations, which adds more noise to the patch. Is it necessary to bundle two distinct changes into a single PR?

fmf/base.py
```diff
@@ -116,88 +157,92 @@ def commit(self):
...
-            return
-        # Attempt to apply the plus operator
         try:
+            # Nothing to do if key not in parent
+            if key not in data:
+                data[key] = value
+                return
+            # Use the special merge for merging dictionaries
+            data_val = data[key]
+            if type(data_val) == type(value) == dict:
```
happz, line 204: It was the original code, but this is not the pythonic way of checking object classes. `if isinstance(data_val, dict) and isinstance(value, dict)` will work better, and supportes `dict` subclasses - there's no reason to not allow e.g. `OrderedDict` to be the source for a node.

fmf/base.py
```diff
@@ -116,88 +157,92 @@ def commit(self):
...
-                    key, self.name))
-        # Subtract numbers
-        if type(data[key]) == type(value) in [int, float]:
-            data[key] = data[key] - value
-        # Replace matching regular expression with empty string
-        elif type(data[key]) == type(value) == type(""):
-            data[key] = re.sub(value, '', data[key])
-        # Remove given values from the parent list
-        elif type(data[key]) == type(value) == list:
-            data[key] = [item for item in data[key] if item not in value]
```
happz, line 190: Yeah, this `type(...) == type(...)` seems to be common - it's definitely a wrong way, and since you're moving those lines around, and it is definitely related to types of values, I'd propose replacing these tests with proper `isinstance()` calls.

fmf/base.py
LecrisUT: Hmm, indeed, I should split it to a separate PR, should it be a standalone PR or bundle a few other style changes (though not sure what to add, maybe converting to `pathlib`?)

fmf/base.py
LecrisUT: What about `str`, do we hard-check `str`?

fmf/base.py
happz: I for one prefer distinct PRs, i.e. one for `format()` vs f-string, another one for `pathlib`, and so on. It's easier for me to review (and revert...) such changes. It's of course possible to create nice, distinct commits focusing one one thing only, and put more of them into a single PR, but that's often a pain when rebasing. For me, one PR for one change wins.

fmf/base.py
LecrisUT: Yes, in those kind of situations. I think it is better to hard-check for `str` type in those cases, and implement a base type if we want other merge-able string subtypes. Dict one, could also get a similar treatment eventually, but I can see a benefit for changing the type there for the time being (i.e. before converting all raw dictionary types to `attrs` dataclasses that would include the `adjust` and other `fmf` directives).

But on that note, do you have a small snippet to confirm that `isinstance` works with `OrderedDict` and `Mapping` types? I don't think it would work with the latter.

happz: > Yes, in those kind of situations. I think it is better to hard-check for `str` type in those cases, and implement a base type if we want other merge-able string subtypes. Dict one, could also get a similar treatment eventually, but I can see a benefit for changing the type there for the time being (i.e. before converting all raw dictionary types to `attrs` dataclasses that would include the `adjust` and other `fmf` directives).

IUIC, a hard check would be `type(value) == type('')`, but what would be the benefit over `isinstance(value, str)`? I'm not sure we understand each other :)

> But on that note, do you have a small snippet to confirm that `isinstance` works with `OrderedDict` and `Mapping` types? I don't think it would work with the latter.

`isinstance(OrderedDict(), dict)` is true, `isinstance(AMappingSubclass(), dict)` is not. `isinstance(AMappingSubclass(), (dict, collections.abc.Mapping))` should cover this base class as well.

LecrisUT: > IUIC, a hard check would be `type(value) == type('')`, but what would be the benefit over `isinstance(value, str)`? I'm not sure we understand each other :)

Yes, though I prefer `type(value) == str`. That would be to enforce we are only checking for `str`. Mostly as a sanity check since I don't think we should be calling `_mergeplus()` as `data[key] += value` in those cases, but instead check for specialized overloads.

> `isinstance(OrderedDict(), dict)` is true, `isinstance(AMappingSubclass(), dict)` is not. `isinstance(AMappingSubclass(), (dict, collections.abc.Mapping))` should cover this base class as well.

Hmm, how about if I implement the general case for `isinstance(value)`, but keep `type(data_val) == dict` for sanity check until we implement `attrs`?

happz: > > IUIC, a hard check would be `type(value) == type('')`, but what would be the benefit over `isinstance(value, str)`? I'm not sure we understand each other :)
> 
> Yes, though I prefer `type(value) == str`. That would be to enforce we are only checking for `str`. Mostly as a sanity check since I don't think we should be calling `_mergeplus()` as `data[key] += value` in those cases, but instead check for specialized overloads.
> > `isinstance(OrderedDict(), dict)` is true, `isinstance(AMappingSubclass(), dict)` is not. `isinstance(AMappingSubclass(), (dict, collections.abc.Mapping))` should cover this base class as well.
> 
> Hmm, how about if I implement the general case for `isinstance(value)`, but keep `type(data_val) == dict` for sanity check until we implement `attrs`?

But that's what I don't understand: why should we enforce `str`, or `dict`, exactly? :) Why bother? As long as `value` is an instance of `str` (or one of its subclasses), it is string-like enough for our use. We can measure its length, we can call its `join()`, and we can add it to another string with `+=`.

I see no benefit in sticking to the exact type only, TBH. From my POV, if `isinstance(value, (dict, Mapping))` is true, then `value` is simply a dictionary, maybe in a trenchcoat, but it still can be iterated over. We perform no black magic with `dict` internals - we copy keys and values between dictionaries, and from time to time we call `get()`, that's all :) It's actually quite common, BTW, e.g. depending on the parser type, `ruamel.yaml` may materialize a YAML file as a pile of `OrderedDict` instances - by allowing `dict` only, we would disqualify otherwise perfectly fine `dict`-like objects. If it walks like a `dict`, if it quacks like a `dict`, then we can merge it with another `dict`-like object.

The same applies to `list` checks, right? We don't inspect `list` type internals, all we do is stuff like `data[key] = [item for item in data[key] if item not in value]` - as long as `value` is list-like, it should be absolutely acceptable.

I don't understand what issues would be solved by allowing the exact type only, or what issues would it prevent. On the other hand, checking for being list-like, dict-like, or string-like opens the door to a wider array of **compatible** input types.

LecrisUT: About `data_val`, it will always be a dict/list/str, etc., unless we open the yaml to beyond `safe_loader`, at which point ambiguity enters (e.g. the original type might no longer be a dict due to a yaml tag, but with `fmf` we want to add to the constructor). If we keep it strict on `data_val`, we would at least catch any issues if something else altered the yaml loader or anything else.

For `value` indeed they have the same interface so the operations will allow us to do the operations. It's just that we are assuming it is what the author of the dict-like, str-like object intends. I think I have an example of a possible confusion in this same PR.

Consider `Tree`. It has a dict-like structure through `__getitem__`, `get()`, etc., but how I have it implemented is that `keys`, `values`, etc. are wrappers to `Tree.data`, while additional `child` or nested items are gathered using `__getitem__` and `get()`. This could create a confusion when using `_merge_plus()`.

For the time being though and until there's a plugin system to add more specialized rules for this, it would be helpful to add the `isinstance` checks for `value`. But I'm not sure how to keep track or denote the places where we might want to change it to have more specialized rules. Due to type-hinting there are quite alot of `isinstance` that we would have to filter through.

happz: > Consider `Tree`. It has a dict-like structure through `__getitem__`, `get()`, etc., but how I have it implemented is that `keys`, `values`, etc. are wrappers to `Tree.data` (because it is easier to implement and because `child` is more natural to access using file-api), while additional `child` or nested items are gathered using `__getitem__` and `get()`. This could create a confusion when using `_merge_plus()`.

I think this example goes further than I wanted: yes, `Tree` might look similar to a dictionary, it does have some methods with similar semantics, but it's not a `dict` subclass. Therefore it would not pass the `isinstance(value, dict)` test, and our code would not try to combine it with other values.

My point was to allow subclasses of `dict` or `list`, e.g. `ruamel.yaml.comments.CommentedMap`, not classes that look like `dict` or `list`, i.e. class tree, not class API. To replace `type(data_val) == type(value) == dict` with `isinstance(data_val, dict) and isinstance(value, dict)` - this would not open the door to dict-like classes like `Tree`. If by any chance a `Tree` instance would be passed to `_merge_plus()`, I would expect none of these tests to match, and such a `value` would be reported as something fmf does not know how to merge with the existing tree.

LecrisUT: What about `isinstance(value, (dict, Mapping))`? I think it's useful for `Tree` to have type-hint of Mapping for further down the line, e.g. if it's used in a generic map input.

PR #205, opened 2023-08-10
Use Click for cli

LecrisUT:
Depends on: #202, #203


LecrisUT:
Took some effort, but I've decoupled this one from the type-hints PR, so that it can be addressed sooner

PR #211, opened 2023-09-15
Fix missing file require for unit tests

KwisatzHaderach:


tests/unit/main.fmf
```diff
@@ -5,4 +5,6 @@ test: python3 -m pytest -v
 framework: shell
 require+:
   - python3-pytest
+  - type: file
+    pattern: examples/*
```
psss, line 9: Shouldn't the pattern start with a slash?

KwisatzHaderach: doesn't really matter, it could be a problem if someone later adds another examples directory somewhere else in this repo it would also be included, but that would happen with the begging slash anyway

psss (approved):

PR #219, opened 2024-01-22
Allow filtering based on the node name

psss:
Add support for filtering based on the node name by allowing plain literals without the `key: name` format.

### DIFF
--- a/fmf/utils.py
+++ b/fmf/utils.py
@@ -211,9 +211,9 @@ def evaluate(expression, data, _node=None):
         raise FilterError("Internal key is not defined: {}".format(error))
 
 
-def filter(filter, data, sensitive=True, regexp=False):
+def filter(filter, data, sensitive=True, regexp=False, name=None):
     """
-    Return true if provided filter matches given dictionary of values
+    Apply advanced filter on the provided data dictionary
 
     Filter supports disjunctive normal form with '|' used for OR, '&'
     for AND and '-' for negation. Individual values are prefixed with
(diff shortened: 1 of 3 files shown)


fmf/utils.py
```diff
@@ -304,13 +306,18 @@ def check_clause(clause):
     if not isinstance(data, dict):
         raise FilterError("Invalid data type '{0}'".format(type(data)))
 
-    # Make sure that data dictionary contains lists of strings
+    # Add fmf name to the data dictionary if provided
     data = copy.deepcopy(data)
+    if name is not None:
```
lukaszachy, line 311: Could we use same "name" for attribute and injected data? eg. 'fmf-name' for both.

psss:
Summary from the hacking session: Use  no key for search by fmf node name. Example:

    /tests/core & tag:quick

So we don't have to invent/decide any name ;-)

lukaszachy:
@psss Is the "/tests/core & tag:quick" possible with current PR? It doesn't seem so.
Also adding a unit test and similar example do the pydoc fmf.filter will be useful.

The-Mule:
This actually brings some issues. See, right now fmf filter is unable parse regular expressions that are quite common and that I assume people will likely be using (there is already [a misplaced issue in tmt project](https://github.com/teemtee/tmt/issues/2633) for this. This problem is solvable[*] under the assumption that keys are always used. But once this PR is merged this assumption is gone.

[*] Using a positive lookahead assertion as follows:

```
# At least one clause must be true
    return any([check_clause(clause)
                for clause in re.split(r"\s*\|\s*(?=[^|]*:)", filter)])
```

Once you have | in a regexp the only way to distinguish it from proper disjunction is to to check that some key follows it. If not then it must be a part of value of the last key. Same approach can be applied to identifying a proper conjunction. With that you can already use pretty powerful regular expressions as values (you still need to avoid using both | and : in you regexp, but probability of that is fairly low IMO). Allowing not to have a key will break this logic.

lukaszachy:
> Once you have | in a regexp the only way to distinguish it from proper disjunction is to to check that some key follows it. If not then it must be a part of value of the last key. Same approach can be applied to identifying a proper conjunction. With that you can already use pretty powerful regular expressions as values (you still need to avoid using both | and : in you regexp, but probability of that is fairly low IMO). Allowing not to have a key will break this logic.

With escaping it is clear whether `|` is part of filter syntax or regexp (`\|`).  Escaping of ':' is already supported (driven by colon used in module nsv strings used as component value), escaping '&' and '|' are planned in this release.

martinhoyer (approved):

fmf/utils.py
```diff
@@ -280,17 +299,41 @@ def check_dimension(dimension, values):
...
+            return bool(re.search(pattern, name))
+        else:
+            return pattern == name
+
     def check_clause(clause):
         """ Split into literals and check whether all match """
         # E.g. clause = 'tag: A, B & tag: C & tag: -D'
         # Split into individual literals by dimension
         literals = dict()
         for literal in re.split(r"\s*&\s*", clause):
```
martinhoyer, line 325: I'm trying to understand how is the escaping of the `&` and `|` supposed to work in the end. Can't we just require the AND and OR operators to have whitespaces around them?
```suggestion
        for literal in [part.strip() for subpart in
               clause.split(' & ') for part in subpart.split(' | ')]:
```

martinhoyer: the splits in list comprehension seems to also take only half as much time as re.split btw.

martinhoyer: L361 would be 
```
    return any([check_clause(clause)
                for clause in filter.split(' | ')])
```

psss: > I'm trying to understand how is the escaping of the `&` and `|` supposed to work in the end. Can't we just require the AND and OR operators to have whitespaces around them?

The plan is to use the backslash character `\` to escape them. The white space is ignored, this is the documeneted behaviour and has been working like this for some time already, so it would not be a good idea to change this as it would break existing configs. Let's continue discussion in #220.

PR #220, opened 2024-01-23
Support escaping `|` and `&` in `utils.filter()`

psss:
Add support for escaping boolean operators using the back slash character `\` to allow for example regular expression grouping.

Pull Request Checklist

* [x] implement the feature
* [x] write the documentation

### DIFF
--- a/fmf/utils.py
+++ b/fmf/utils.py
@@ -222,6 +222,13 @@ def filter(filter, data, sensitive=True, regexp=False, name=None):
         tag: Tier1 | tag: Tier2 | tag: Tier3
         category: Sanity, Security & tag: -destructive
 
+    Use the back slash character ``\\`` to escape the boolean
+    operators if you need to specify them as part of the filter
+    expressions::
+
+        tag: Tier(1\\|2)
+        text: Q\\&A
+
     Note that multiple comma-separated values can be used as a syntactic
     sugar to shorten the filter notation::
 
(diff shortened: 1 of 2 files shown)


lukaszachy (commented):
Could you add examples how this should be used? I'm not sure the re.sub is correct but hard to tell.

The-Mule:
@lukaszachy Any regexp where `|` is quite common, e.g. `tmt test ls --filter 'name: (/setup\|/functional).*'`.

lukaszachy:
So `re.sub` doesn't work as expected, or? I'd assume that it intention is to 'undo' the escaping.

BTW it would be really good to see this covered in unit tests...

```
import re

>>> clause = re.split(r"\s*(?<!\\)\|\s*", r'foo\|bar|baz')[0]
>>> clause
'foo\\|bar'
>>> re.sub(r"\|", "|", clause)
'foo\\|bar'
```

psss:
It's split of the `filter()` clause versus regular expression. See my [comment](https://github.com/teemtee/tmt/issues/2633#issuecomment-2107513200) in the related issue reported by @kkaarreell.

psss:
Let's include escaping of `&` as well.

martinhoyer:
Found this PR only after commenting at https://github.com/teemtee/fmf/pull/219#discussion_r1624484311

lukaszachy (approved):

PR #223, opened 2024-02-27
Add directive `select` to force selecting node

lukaszachy:
- [x] implement the feature
- [x] write the documentation
- [x] extend the test coverage


lukaszachy:
Limiting tests in #225. Lets see

jscotka:
Hi, I would love some more descriptive name, instead of `is-leaf` e.g. someghing like `marked-as_leaf`, or I can imagine something underscoder, to avoid confusion user data and TMT will then ignore these attributes, like:
`_mark_as_leaf: True` and in TMT there could be rule to skip underscored items, as they are directives and should not be used as metadata source directly.

happz:
> ... or I can imagine something underscoder, to avoid confusion user data and TMT will then ignore validation these attributes, like: `_mark_as_leaf: True` and in TMT there could be rule to skip underscored items

tmt can easily ignore directives with dashes simply by listing them in tmt sources, if needed, plus users may define their custom keys with underscores in their tests (`extra-foo_bar`, https://github.com/teemtee/tmt/blob/main/tmt/schemas/test.yaml#L128), therefore a rule as general as "ignore everything with underscores" might cause more harm than good.

I suspect these directives don't even reach tmt, do they? They are consumed by fmf, applied and tmt gets the final materialized tree of data, correct?

LecrisUT:
What about also having the opposite? I.e. marking a `file.fmf` as non-leaf. I was thinking if this might be necessary in the context of:
https://github.com/teemtee/fmf/blob/29190c78e4065adf98cbb3ccf747b01973d761e2/fmf/base.py#L812-L815

---

More specifically, in https://github.com/LecrisUT/tmt-copr/issues/2 I want `discover` plugin to inject  `main.fmf`-like file into:
```yaml
/prefix:
  framework: copr
  test: custom-placeholder
/prefix/pkg1:
  package: pkg1
```
Then there could be overrides defined in `/prefix/pkg1.fmf` like
```yaml
result: xfail
```

The idea of adding a function to mark something as non-leaf is if we only have a `/prefix.fmf` file:
```yaml
duration: 10h
```
In this case `tmt lint` will fail because `/prefix.fmf` is neither a plan, nor a test. It would be nice to mark this as being non-leaf so that it is not parsed by `tmt lint` (or at least not before `discover` is run) or other phases.

lukaszachy:
> I suspect these directives don't even reach tmt, do they? They are consumed by fmf, applied and tmt gets the final materialized tree of data, correct?

Correct.

jscotka:
> > I suspect these directives don't even reach tmt, do they? They are consumed by fmf, applied and tmt gets the final materialized tree of data, correct?
> 
> Correct.

This is question. On first singht, I can imagine that TMT will interpret the value, but another is, that it is semantic of FMF, so that it should be materialized and TMT will do not see this key anyway.
as @happz mentioned, the underscored semantics could cause harm in case of ignoring them. In opposite side, when someone using is-leaf attribute today. it will break someone workflow in future, so that Yes, This underscoring is not good, But I would like to see also some special mark to intepret these values, e.g. long time ago, we've discussed of using `./:` or something similar as node name path.
so that I can imagine also sometghing like:
```yaml
this_is_leaf/:
  ./:
    mark_as_leaf: True
    another_fmf_firective: something
  test: shell.sh
````

BTW, this disussion is closer to FMF/TMT config/profiles (https://github.com/teemtee/fmf/pull/196) than we should allow :-), it allows you to configure/override default FMF behaviour, what do you think? From one perspective it is very simple PR, from another perspective overriding, redefining nodes or change some behaviour seems to me very close to profiles and configs, that's Why I would like to see something more generic, than just one attribute to change befaviour of FMF, but do it more extensible.

**Comment by lukaszachy** [comment]
> This is question. On first singht, I can imagine that TMT will interpret the value, but another is, that it is semantic of FMF, so that it should be materialized and TMT will do not see this key anyway.

tmt  doesn't see this value, similar to merge suffixes (+, -, +<), all of this is processed by fmf.

**Comment by lukaszachy** [comment]
> BTW, this disussion is closer to FMF/TMT config/profiles (#196) than we should allow :-), it allows you to configure/override default FMF behaviour,

I beg do differ - for me this is just setting fmf directive. By overriding default fmf behavour I'd expect custom _merge_special, injection of directives not known to fmf or similar.

**Comment by lukaszachy** [comment]
Hm, I'll leave the `mark-non-leaf` feature to the future. As it changes the flow of creating the tree a lot -- if I understood correctly such nodes will be effectively 'main.fmf' -> not sure what to do if more are present, which data win...

**Comment by lukaszachy** [comment]
Could you please create a new issue, @LecrisUT ?

What I have in mind is this:
```
.
├── bar
│   └── main.fmf
├── .fmf
│   └── version
├── foo.fmf
└── main
```

Currently this produces leaf /, non-leaves /foo and /bar. Inheritance works in way that /bar is created based on 'main.fmf' and 'bar/main.fmf'.

If foo.fmf is 'mark-non-leaf' - how it will ever become a leaf aka node usable by tmt? 
Should it become parent of some none, similar to the 'main.fmf'?

**Comment by LecrisUT** [comment]
> Currently this produces leaf /, non-leaves /foo and /bar.

You mean the opposite right? non-leaf `/` and `/foo`, `/bar` are leaves (in the current implementation, and discussion does not touch the current implementation right?)

> If foo.fmf is 'mark-non-leaf' - how it will ever become a leaf aka node usable by tmt?
Should it become parent of some none, similar to the 'main.fmf'?

The idea is to support the inverse of grafting where the base is populated by `discover` and on top of that we apply `fmf.Tree` from filesystem:
- grafting: have a pre-defined `fmf.Tree` from filesystem then add on top of that another `fmf.Tree` (from dict usually)
- inverse-grafting: construct an `fmf.Tree`-like structure, then apply `fmf.Tree` from filesystem

But I guess in both cases you could have breaking tmt lint due to incomplete data, so maybe it only needs to be on the `tmt` side (other than `graft`/`inverse_graft` support on `fmf` side)

**Comment by lukaszachy** [comment]
> You mean the opposite right? non-leaf `/` and `/foo`, `/bar` are leaves (in the current implementation, and discussion does not touch the current implementation right?)

Right you are, I flipped those two terms.

File: docs/features.rst [hunk 1]
```diff
@@ -194,6 +194,14 @@ The key content attributes are not supposed to be hard-coded in
 the Flexible Metadata Format but freely configurable. Multiple key
 content attributes (e.g. script & backend) could be used as well.
 
+Sometimes it is necessary to select node from the metadata tree
+even though it is not leaf. For example, when virtual tests are
+created from a parent test but one wants to keep the parent available
+as a test as well. To do so, one can set fhe following directive::
+
+    /:
+        make-leaf: true
```
psss, line 203: This approach seems a bit one-use-case focused. What about making the syntax a little bit more flexible and extensible? It could look like this:

    /:
        kind: leaf

This new directive would denote `kind` of the node, or `type`, or some better name if we find one. By default leaves and branches would be detected from the presence of children but it could be explicitly overridden by user if desired. Possible values could be:

* branch
* leaf
* hidden

When `climb()`ing the tree, it would be possible to select whether leaves or branches should be yielded and whether hidden nodes should be ignored or not. What do you think?

psss: Alternative approach from the hacking session:

 * show: boolean

By default all leaves would have `show: true` and branches `show: false`, but additional branches could be also assigned `show: true` or leaves hidden by `show: false`.

lukaszachy: @LecrisUT Will this changed approach work for you (and https://github.com/teemtee/fmf/issues/231) ?

LecrisUT: Indeed that's a cleaner solution. Only comment is if `show` is intuitive that it affects evaluation and not just `fmf show`/`tmt plans show`.

lukaszachy: Good point, while writing docs I've realized that `select` is probably more direct name.

lukaszachy:
PR rebased, in the end I went with 'select' as directive name as it made more sense in the docs.

docs/features.rst
LecrisUT: It is not immediately apparent, but after swirling it around in the head, it indeed becomes intuitive. I think it's just like the `inherit`, operators, etc., once you get in the mindset (what's the subject, result, etc.) it becomes intuitive

psss (approved):
Nicely done! Thanks.

martinhoyer (approved):
Cool!

docs/features.rst
psss: Yeah, agreed, `select` sounds good.

docs/features.rst
```diff
@@ -194,6 +194,20 @@ The key content attributes are not supposed to be hard-coded in
...
 content attributes (e.g. script & backend) could be used as well.
 
+Select
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Sometimes it is necessary to select node from the metadata tree
+even though it is not a leaf. For example, when virtual tests are
+created from a parent test but one wants to keep the parent available
+as a test as well. On the other hand, one might want to hide leaf node,
+instead of deleting it completely. To do so, one can set fhe directive::
```
LecrisUT, line 204:
```suggestion
instead of deleting it completely. To do so, one can set the directive::
```

LecrisUT (approved):

PR #225, opened 2024-02-28
Limit number of tests run for integration with tmt

lukaszachy:
Just Tier:1 and only with 'local' provision.
We need to check fmf format works, not tmt features...


psss (approved):
Yeah, makes sense, +1.

PR #229, opened 2024-04-11
Avoid `copy()` calls in `adjust()`

lukaszachy:
Replace `.pop()` calls with skipping "control" keys (when,because..) during the merge.

### DIFF
--- a/fmf/base.py
+++ b/fmf/base.py
@@ -23,6 +23,7 @@ from fmf.utils import dict_to_yaml, log
 SUFFIX = ".fmf"
 MAIN = "main" + SUFFIX
 IGNORED_DIRECTORIES = ['/dev', '/proc', '/sys']
+ADJUST_CONTROL_KEYS = ['because', 'continue', 'when']
 
 
 # ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
@@ -373,7 +374,7 @@ class Tree:
 
         # Adjust rules should be a dictionary or a list of dictionaries
         try:
-            rules = copy.deepcopy(self.data[key])
+            rules = self.data[key]
             log.debug("Applying adjust rules for '{}'.".format(self))
             log.data(rules)
             if isinstance(rules, dict):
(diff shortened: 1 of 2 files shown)


fmf/base.py
```diff
@@ -220,6 +221,9 @@ def _merge_minus(self, data, key, value):
     def _merge_special(self, data, source):
         """ Merge source dict into data, handle special suffixes """
         for key, value in sorted(source.items()):
+            # Merge only attribute keys, not the adjust control keys
+            if key in ADJUST_CONTROL_KEYS:
+                continue
```
psss, line 226: The `_merge_special()` method is used in `Tree.merge()` as well where child data are merged with parent data so this would break regular metadata using one of the reserved keys. For example:

```yaml
what: something

/child:
    when: yesterday
```

Before the change we get:

    > fmf show
    /child
    what: something
    when: yesterday

But after the child value is not included.

    > fmf show
    /child
    what: something

psss (changes requested):
Thanks for the improvement. The simplification looks good. But there is a regression. We should probably apply the `ADJUST_CONTROL_KEYS` condition only for selected `_merge_special()` calls. What about including a special parameter which which turn the condition on and off?

fmf/base.py
lukaszachy: ah. thanks for catching this.

psss: Or, what about just filtering the extra keys when calling the method in the `adjust` case?

```python
self._merge_special(
    self.data, {
        key: value
        for key, value in rule.items()
        if key not in ADJUST_CONTROL_KEYS
        }
    )
```

lukaszachy: fixed in 9cef668

martinhoyer (commented):
@lukaszachy Could you please provide a usage example, to help me understand the context? :pray:   

EDIT: Wait, wrong PR. I meant for https://github.com/teemtee/fmf/pull/230

fmf/base.py
```diff
@@ -24,6 +24,7 @@
 SUFFIX = ".fmf"
 MAIN = "main" + SUFFIX
 IGNORED_DIRECTORIES = ['/dev', '/proc', '/sys']
+ADJUST_CONTROL_KEYS = ['because', 'continue', 'when']
```
martinhoyer, line 27: Perhaps this should be a set instead of a list?

martinhoyer (approved):
Nice.

fmf/base.py
psss: Could be, but `list` is ok as well, I'd say. Let's not block on this.

fmf/base.py
psss: Good, thanks!

psss (approved):

PR #230, opened 2024-04-11
Accept additional rules in the `adjust()` call

lukaszachy:
Use case: Explicit set of the rules to be applied, regardless what is in  the Tree.

Could be simplified if/once #229 is merged.

### DIFF
--- a/fmf/base.py
+++ b/fmf/base.py
@@ -368,9 +368,9 @@ class Tree:
         adjust rule, and whether it was applied or not.
 
         Optional 'additional_rules' parameter can be used to specify rules
-        which should be applied in addition to those defined in the node itself.
-        These additional rules are processed even when the applied
-        rule defined in the data has ``continue: false`` set.
+        that should be applied after those from the node itself.
+        These additional rules are processed even when an applied
+        rule defined in the node has ``continue: false`` set.
         """
 
         # Check context sanity
--- a/fmf/base.py
+++ b/fmf/base.py
@@ -5,6 +5,7 @@ import os
 import re
 import subprocess
 from io import open
+from itertools import chain
 from pprint import pformat as pretty
 from typing import Any, Dict, Optional, Protocol
 
@@ -343,7 +344,8 @@ class Tree:
         log.data(pretty(self.data))
 
     def adjust(self, context, key='adjust', undecided='skip',
-               case_sensitive=True, decision_callback=None):
+               case_sensitive=True, decision_callback=None,
+               additional_rules=None):
         """
         Adjust tree data based on provided context and rules
 
@@ -365,6 +367,11 @@ class Tree:
         Optional 'decision_callback' callback would be called for every adjust
         rule inspected, with three arguments: current fmf node, current
         adjust rule, and whether it was applied or not.
+
+        Optional 'additional_rules' parameter can be used to specify rules
+        which should be applied in addition to those defined in the node itself.
+        These additional rules are processed even when the applied
+        rule defined in the data has ``continue: false`` set.
         """
 
         # Check context sanity
@@ -387,10 +394,18 @@ class Tree:
         except KeyError:
             rules = []
 
+        additional_rules = copy.deepcopy(additional_rules)
+
+        # Accept same type as rules from data
+        if additional_rules is None:
+            additional_rules = []
+        elif isinstance(additional_rules, dict):
+            additional_rules = [additional_rules]
+
         context.case_sensitive = case_sensitive
 
         # Check and apply each rule
-        for rule in rules:
+        for rule in chain(rules, additional_rules):
 
             # Rule must be a dictionary
             if not isinstance(rule, dict):
@@ -423,7 +438,7 @@ class Tree:
                         }
                     self._merge_special(self.data, apply_rule)
 
-                    # First matching rule wins, skip the rest unless continue
+                    # First matching rule wins, skip the rest of this set unless continue
                     if not continue_:
                         break
                 else:
@@ -446,7 +461,9 @@ class Tree:
         # Adjust all child nodes as well
         for child in self.children.values():
             child.adjust(context, key, undecided,
-                         case_sensitive=case_sensitive, decision_callback=decision_callback)
+                         case_sensitive=case_sensitive,
+                         decision_callback=decision_callback,
+                         additional_rules=additional_rules)
 
     def get(self, name=None, default=None):
         """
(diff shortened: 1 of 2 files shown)


fmf/base.py
```diff
@@ -363,6 +365,10 @@ class describing the environment context. By default, the key
         Optional 'decision_callback' callback would be called for every adjust
         rule inspected, with three arguments: current fmf node, current
         adjust rule, and whether it was applied or not.
+
+        Optional 'additional_rules' parameter can be used to specify rules
+        which should be apple in addition to those defined in the data themselves.
```
happz, line 370: `s/apple/applied/`

lukaszachy: reworded, thanks for spotting this typo

fmf/base.py
```diff
@@ -385,65 +391,75 @@ class describing the environment context. By default, the key
...
+                        raise utils.GeneralError(
+                            "Invalid value for the 'undecided' parameter. Should "
+                            "be 'skip' or 'raise', got '{}'.".format(undecided))
 
         # Adjust all child nodes as well
         for child in self.children.values():
             child.adjust(context, key, undecided,
-                         case_sensitive=case_sensitive, decision_callback=decision_callback)
+                         case_sensitive=case_sensitive, decision_callback=decision_callback,
+                         additional_rules=additional_rules)
```
martinhoyer, line 462:
```suggestion
                         case_sensitive=case_sensitive, 
                         decision_callback=decision_callback,
                         additional_rules=additional_rules)
```

fmf/base.py
```diff
@@ -363,6 +364,11 @@ class describing the environment context. By default, the key
         Optional 'decision_callback' callback would be called for every adjust
         rule inspected, with three arguments: current fmf node, current
         adjust rule, and whether it was applied or not.
+
+        Optional 'additional_rules' parameter can be used to specify rules
+        which should be applied in addition to those defined in the data
+        themselves. These additional rules are processed even when the applied
```
martinhoyer, line 370: Not sue I have the context, but `in the data themselves` sounds odd to me
```suggestion
        which should be applied in addition to those defined in the data
        itself. These additional rules are processed even when the applied
```

fmf/base.py
```diff
@@ -385,65 +391,75 @@ class describing the environment context. By default, the key
...
-                    self._merge_special(self.data, rule)
+        context.case_sensitive = case_sensitive
 
-                    # First matching rule wins, skip the rest unless continue
-                    if not continue_:
-                        break
-                else:
+        # Check and apply each rule
+        for set_of_rules in [rules, additional_rules]:
+            for rule in set_of_rules:
```
martinhoyer, line 406: How about doing
```
set_of_rules = rules + additional_rules
for rule in set_of_rules:
```
to avoid the extra indentation?

psss:
> Could be simplified if/once #229 is merged.

#229 has been merged, @lukaszachy, could you please rebase?

fmf/base.py
lukaszachy: google search about data is/are was fun. But en_US seems to prefer 'is'.

fmf/base.py
lukaszachy: good point, it made rebase much easier.

fmf/base.py
lukaszachy: But in the end I reworded it, hopefully it now makes more sense

fmf/base.py
```diff
@@ -365,6 +367,11 @@ class describing the environment context. By default, the key
         Optional 'decision_callback' callback would be called for every adjust
         rule inspected, with three arguments: current fmf node, current
         adjust rule, and whether it was applied or not.
+
+        Optional 'additional_rules' parameter can be used to specify rules
+        which should be applied in addition to those defined in the node itself.
+        These additional rules are processed even when the applied
+        rule defined in the data has ``continue: false`` set.
```
LecrisUT, line 374: About the `continue` comment, isn't that contradicting with the `if not continue_: break`?

psss: Yeah, it seems that if any of the node rules has `continue: false` the loop is ended an thus additional rules would not be processed.

psss (changes requested):
Looks good, except for the `continue` statement.

fmf/base.py
lukaszachy: good catch. 'continue: false' needs to be bound to the 'node' or 'additional_rules'

lukaszachy: rationale: 'additional_rules' comes from runtime and they _have to be processed regardless 'continue' value in the rules from node metadata.

LecrisUT: For me `additional` reads as append, probably because of the relation to `+`. Was hoping that `+<` had a documented naming to use for the alternative, but no luck.

Had to detangle the concepts of rule and grammar in my head since this seems to address the former. So the usage of it is like adding dict key-values to the `adjust` (although we have the issue of inconsistent operation order) in which case, similar to inheritance it feels like it should be appended. (For the grammar part of the operations, it shouldn't really matter, which is why I got confused for a second)

fmf/base.py
lukaszachy: restored, added note. Is there a nicer way? All I could think of is to move the inner loop into own function.

martinhoyer: Perhaps https://docs.python.org/3.9/library/itertools.html#itertools.chain

fmf/base.py
lukaszachy: Described in 886fcdb

lukaszachy: The only motivation for this is to make https://github.com/teemtee/tmt/discussions/2430.

fmf/base.py
LecrisUT: Doesn't read like it would survive a `break`. I think it's more for maintaining the generator aspect of the iterator.

Would feel very error prone to have any mechanism that would interfere with a language specific like `break`. I think this approach is readable enough. Maybe another approach would be to manipulate the iterators instead of the break, but that would not read well.

fmf/base.py
psss: Thanks for the clarification and docs enhancement.

psss (approved):
Now looks good, thanks!

fmf/base.py
psss: Agreed. Yeah, this is nicely readable.

fmf/base.py
psss: Ack, it's better.

PR #234, opened 2024-04-18
Implement the `~` merge operation

lukaszachy:
WDYT?

Todo:
- [x] find BETTER_NAME for this helper function
- [x] Documentation
- [x] smuggle '-~' operator (delete if matches regexp)

### DIFF
--- a/docs/features.rst
+++ b/docs/features.rst
@@ -121,6 +121,41 @@ attribute, matching regular expressions are replaced by an empty
 string. For dictionaries it's possible to provide list of keys
 which should be removed.
 
+Substitution of current values can be done by appending a ``~``
+suffix to the key name. The pattern and replacement parameters
+need to be provided as values in the form of
+``<d>PATTERN<d>REPLACEMENT<d>``, where ``<d>`` is delimiter which
+can be any character however such character cannot be then used
+within PATTERN and REPLACEMENT text as escaping isn't supported.
+This input can be either a string or list of strings.
+
+The `re.sub`__ is used to do the substitution thus all features of
+``re.Pattern`` can be used (named groups, back referencing...).
+
+In the fmf file it is better to use single quotes ``'`` as they do
+not need such intensive escaping::
+
+    require~: ';^foo;foo-ng;'
+    recommend~:
+      - '/python2-/python3-/'
+
+__ https://docs.python.org/3/library/re.html#re.sub
+
+Remove parent value only if it matches regular expression is done
+using  the ``-~`` suffix. If value matches any of provided
+`regular expressions`__ it is removed. If the parent value is a
+list, the matching item is removed from this list. If the parent
+value is a string, the value is set to an empty string. If the
+parent value is a dictionary, the matching key is removed. These
+regular expressions can be just a single item or a list of
+strings::
+
+    description-~: '.*'
+    require-~:
+      - 'python2.*'
+
+__ https://docs.python.org/3/library/re.html#regular-expression-syntax
+
 
 Elasticity
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/examples/merge/parent.fmf
+++ b/examples/merge/parent.fmf
@@ -36,3 +36,18 @@ very:
                 x+: 1
                 y-: 1
                 z: 0
+
+# Regexp used to change values - note that the second tag regexp cannot be applied
+# as Tier2 is already modified into 't2'
+/regexp:
+    description~: "/.en(.*) .*/gen\\1/"
+    tags~:
+    - '/Tier(.)/t\1/'
+    - /Tier2/t3/
+
+/minus-regexp:
+    description-~: '.*'
+    tags-~:
+    - '.ier1'
+    vars-~:
+    - 'y'
--- a/fmf/base.py
+++ b/fmf/base.py
@@ -191,6 +191,49 @@ class Tree:
                 "MergeError: Key '{0}' in {1} ({2}).".format(
                     key, self.name, str(error)))
 
+    def _merge_regexp(self, data, key, value):
+        """ Handle substitution of current values """
+        if isinstance(value, str):
+            value = [value]
+        for pattern, replacement in [utils.split_pattern_replacement(v) for v in value]:
+            if isinstance(data[key], list):
+                try:
+                    data[key] = [re.sub(pattern, replacement, original) for original in data[key]]
+                except TypeError:
+                    raise utils.MergeError(
+                        "MergeError: Key '{0}' in {1} (not a string).".format(
+                            key, self.name))
+            elif isinstance(data[key], str):
+                data[key] = re.sub(pattern, replacement, data[key])
+            else:
+                raise utils.MergeError(
+                    "MergeError: Key '{0}' in {1} (wrong type).".format(
+                        key, self.name))
+
+    def _merge_minus_regexp(self, data, key, value):
+        """ Handle removing current values if they match regexp """
+        # A bit faster but essentially `any`
+        def lazy_any_search(item, patterns):
+            for p in patterns:
+                if re.search(p, str(item)):
+                    return True
+            return False
+        if isinstance(value, str):
+            value = [value]
+        if isinstance(data[key], list):
+            data[key] = [item for item in data[key] if not lazy_any_search(item, value)]
+        elif isinstance(data[key], str):
+            if lazy_any_search(data[key], value):
+                data[key] = ''
+        elif isinstance(data[key], dict):
+            for k in list(data[key].keys()):
+                if lazy_any_search(k, value):
+                    data[key].pop(k)
+        else:
+            raise utils.MergeError(
+                "MergeError: Key '{0}' in {1} (wrong type).".format(
+                    key, self.name))
+
     def _merge_minus(self, data, key, value):
         """ Handle reducing attributes using the '-' suffix """
         # Cannot reduce attribute if key is not present in parent
@@ -225,8 +268,12 @@ class Tree:
                 self._merge_plus(data, key.rstrip('+'), value)
             elif key.endswith('+<'):
                 self._merge_plus(data, key.rstrip('+<'), value, prepend=True)
+            elif key.endswith('-~'):
+                self._merge_minus_regexp(data, key.rstrip('-~'), value)
             elif key.endswith('-'):
                 self._merge_minus(data, key.rstrip('-'), value)
+            elif key.endswith('~'):
+                self._merge_regexp(data, key.rstrip('~'), value)
             # Otherwise just update the value
             else:
                 data[key] = value
--- a/fmf/utils.py
+++ b/fmf/utils.py
@@ -948,3 +948,37 @@ def validate_data(data, schema, schema_store=None):
             jsonschema.exceptions.UnknownType
             ) as error:
         raise JsonSchemaError(f'Errors found in provided schema: {error}')
+
+
+class PatternReplacement(NamedTuple):
+    pattern: str
+    replacement: str
+
+
+def split_pattern_replacement(source):
+    """
+    Splits pattern/replacement string input into parts
+
+    Format of the input:
+        <delimiter><PATTERN><delimiter><REPLACEMENT><delimiter>
+
+    Delimiter set by the first character of the input and this character
+    cannot be used in the PATTERN or REPLACEMENT text. Escaping is not
+    supported.
+    """
+
+    try:
+        delimiter = source[0]
+        if source[-1] != delimiter:
+            raise FormatError("'{0}' has to end with '{1}'.".format(source, delimiter))
+    except IndexError:
+        raise FormatError("'{0}' has to start and end with the same delimiter.".format(source))
+
+    try:
+        pattern, replacement = source[1:-1].split(delimiter)
+    except ValueError:
+        raise FormatError("'{0}' can't be split in two parts".format(source))
+
+    if not pattern:
+        raise FormatError("Pattern cannot be empty: '{0}'.".format(source))
+    return PatternReplacement(pattern=pattern, replacement=replacement)
--- a/tests/unit/test_utils.py
+++ b/tests/unit/test_utils.py
@@ -483,3 +483,25 @@ class TestDictToYaml:
         data = dict(y=2, x=1)
         assert fmf.utils.dict_to_yaml(data) == "y: 2\nx: 1\n"
         assert fmf.utils.dict_to_yaml(data, sort=True) == "x: 1\ny: 2\n"
+
+
+class TestSplitPatternReplacement:
+    @pytest.mark.parametrize("src", [
+        "",  # Empty input
+        "/",  # Not long enough input
+        ";/;",  # Not long enough - missing 'REPL;'
+        ";;;",  # PATTERN and REPL parts empty
+        "/a/b/c",  # Input after trailing deliminer
+        "/a/b/c/d/",  # More than 3 delimiters
+        "/x;y/",  # No REPL part as delimiter is /
+        "/a/a;"  # No trailing delimiter
+        ])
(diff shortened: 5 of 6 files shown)


fmf/utils.py
```diff
@@ -853,3 +853,26 @@ class JsonSchemaValidationResult(NamedTuple):
...
+
+class PatternRepl(NamedTuple):
+    pattern: str
+    repl: str
+
+
+def BETTER_NAME(source):
+    """ Splits sed-like input into pattern/repl pair suitable for re.sub """
+    try:
+        if source[0] != source[-1] != '/':
```
LecrisUT, line 866: Why not support any character and use `string.split` to avoid processing any escape characters? Than everything would be escaped/processed as normal regex patterns. There might be some trickery needed to escape `source[0]`, but that should have a predictable pattern to find-and-replace with normal string operations before parsing regex: `'\' + escape_char` -> `escape_char`

This would be important to support because escaping `/very/long/path/to/my/data` will be a nightmare to read.

examples/merge/parent.fmf
```diff
@@ -36,3 +36,11 @@ very:
                 x+: 1
                 y-: 1
                 z: 0
+
+# Regexp used to change values - note that the second tag regexp cannot be applied
+# as Tier2 is already modified into 't2'
+/regexp:
+    description~: "/.en(.*) .*/gen\\1/"
+    tags~:
+    - "/Tier(.)/t\\1/"
```
LecrisUT, line 45: Worth documenting the `\\1` vs `\1` pattern, and compare with `sed` and python `regex`. That would also help the reader get an intuition on how this is handled

LecrisUT: My intuition is that except for the main separator character everything would be parsable as native python-regex syntax, e.g. something like this should be parsable `'/(?i)prefix(?P<my_name>[\.\d]+)/replace_pattern/'` (i.e. to be able to toggle case-sensitivity, use named patterns etc.).

The replaced syntax might be tricky and I think `sed` does not allow named regex groups, but python supports it with `\g<my_name>`. Alternatively I was thinking `f-string` format would be intuitive, but maybe hard to implement.

Documentation wise I would say (just talking points, better paragraph should be written):
> The `~` find-and-replace pattern follows a `sed`-like pattern with python-native regex functions:
>
> `<separator><regex_search_pattern><separator><regex_replace_pattern><separator>`
>
> You may choose any `<separator>` character similar to `sed` format. Common patterns are `/`, `#`, `|`. Choose one that best fits your pattern. If necessary you can escape the separator by prepending `\` to it. Any regex pattern processing is only happening after the raw string is split by the separator.
> 
> The search and replace patterns are treated as python's `re.pattern` supporting any operations like `(?i)`, `(?P<named_group>)`, etc., and for the replace pattern you can use constructs like `\g<named_group>`, `\g<0>`

For yaml do we need to use `'` to better handle the string escapes?

fmf/utils.py
lukaszachy: If I don't need to implement escaping, as in: Set the leading character properly as it cannot be used in pattern nor in repl then +1. I didn't think it would be acceptable ;)

examples/merge/parent.fmf
lukaszachy: The `\\1` vs `\1` is just way how to get raw string into python, isn't it? So you are right - it needs to be `"\\1"` but can be `'\1'` (double vs single quotes).

lukaszachy: How about this docs:

The `~` text substitution operator evaluates `re.sub` function on each existing value of the key. The 'pattern' and 'repl' parameters have to be provided in the form of `<separator><pattern value><separator><repl value><separator>`. You may choose any character as the  separator, if such character is used in 'pattern' or 'repl' it will have to be escaped by prepending `\` to it. After the split of the substitution value into the pattern and repl arguments the separator character will be unescaped. Except of this no further processing of the substitution values happen.

See https://docs.python.org/3/library/re.html#re.sub for re.sub capabilities.  All occurrences of the pattern are replaced and pattern flags are not set.

For better readability use single quotes `'` in the fmf file, as they do not require the substitution value to be further escapes to meet the YAML syntax.

fmf/utils.py
lukaszachy: But even with escaping it should be easy to change.

examples/merge/parent.fmf
LecrisUT: > How about this docs:

Yep, the docs seem good. Just one note that `pattern` and `repl` are a bit specialized. It would be good to add something  like `'pattern' (as in search pattern)` and `'repl` (as in replace format)` to be more readable for newcomers.

lukaszachy:
`~` should be done now 

I'll add tests and docs for '-~' tomorrow.

tests/unit/test_utils.py
```diff
@@ -434,3 +434,15 @@ def test_sort(self):
         data = dict(y=2, x=1)
         assert fmf.utils.dict_to_yaml(data) == "y: 2\nx: 1\n"
         assert fmf.utils.dict_to_yaml(data, sort=True) == "x: 1\ny: 2\n"
+
+
+class TestBETTERNAME:
+    def test_invalid(self):
+        for s in ["", "/", ";/;", ";;;", "/a/b/c", "/a/b/c/d",
```
LecrisUT, line 441: Quick comment, how about splitting this list line-by-line so we can add comment on what is caught in each part

lukaszachy: updated

docs/features.rst
```diff
@@ -121,6 +121,33 @@ attribute, matching regular expressions are replaced by an empty
...
+which can be any character however such character cannot be then used withing PATTERN
+and REPLACEMENT text as escaping isn't supported. These input can be either a string or
+list of strings.
+
+The `re.sub`__ is used to do the substitution thus all feaures of ``re.Pattern`` can
+be used (named groups, backreferencing...).
+
+In the fmf file it is better to use single quotes ``'`` as they do not need such intensive escaping::
+
+    require~: ';^foo;foo-ng;'
```
psss, line 136: Just a quick first-sight brainstorm: Instead of describing the delimiters and parsing them and escaping them wouldn't it be easier to have two keys?

```yaml
require~:
    pattern: foo
    replace: foo-ng
```

Or do you find to too long?

lukaszachy: @psss Some details in https://github.com/teemtee/tmt/discussions/2430#discussioncomment-9131018 

I think that fmf prefers shorter way of syntax so "sed"-like seems (IMO) to be a better fit.

LecrisUT: It could be confusing because no other operation sets dict keys. I am thinking of how it fits along with the `+` operator where the value is the operand. It's hard to describe

fmf/base.py
```diff
@@ -191,6 +191,44 @@ def _merge_plus(self, data, key, value, prepend=False):
                 "MergeError: Key '{0}' in {1} ({2}).".format(
                     key, self.name, str(error)))
 
+    def _merge_regexp(self, data, key, value):
+        """ Handle substitution of current values """
+        if isinstance(value, str):
+            value = [value]
+        for pattern, repl in [utils.split_pattern_repl(v) for v in value]:
+            if isinstance(data[key], list):
+                data[key] = [re.sub(pattern, repl, original) for original in data[key]]
```
LecrisUT, line 200: Shouldn't there be a check or try catch that `original` is a `str`?

docs/features.rst
psss: Thanks for the pointer. I've read he discussion. Makes sense to keep the syntax concise.

fmf/base.py
lukaszachy: fixed in 5de1c69

docs/features.rst
```diff
@@ -121,6 +121,33 @@ attribute, matching regular expressions are replaced by an empty
 string. For dictionaries it's possible to provide list of keys
 which should be removed.
 
+Substitution of current values can be done by appending a ``~`` suffix
+to the key name. The pattern and replacement parameters need to be provided
+as values in the form of ``<d>PATTERN<d>REPLACEMENT<d>``, where ``<d>`` is delimiter
+which can be any character however such character cannot be then used withing PATTERN
+and REPLACEMENT text as escaping isn't supported. These input can be either a string or
```
thrix, line 128: typo: `These input` ...

lukaszachy:
> Issue is that I can't. I can only resolve comments on PRs that I have opened.

Yeah. This limitation from github is really strange:/ I have no idea that why anyone who _can_ comment is not suddenly allowed to resolve conversation they started.

docs/features.rst
lukaszachy: fixed 25275a1

lukaszachy:
`/examples/merge/parent.fmf` now uses also single quotes so the `\1` doesn't need to be escaped

tests/unit/test_utils.py
```diff
@@ -434,3 +434,23 @@ def test_sort(self):
...
+
+
+class TestSplitPatternRepl:
+    @pytest.mark.parametrize("src", [
+        "",  # Empty input
+        "/",  # Not long enough input
+        ";/;",  # Not long enough - missing 'REPL;'
+        ";;;",  # PATTERN and REPL parts empty
+        "/a/b/c",  # Input after trailing deliminer
+        "/a/b/c/d",  # More parts after trailig delimiter
```
LecrisUT, line 446:
```suggestion
        "/a/b/c/d/",  # More than 3 delimiters
```

Isn't the original one basically identical to `/a/b/c`? I think you want to test this case instead?

LecrisUT (commented):
Sorry for the late review, just a few minor touches and it's all :+1:. You can resolve all of my previous discussions.

martinhoyer:
I like the functionality, but this might be too much regex magic from my PoV. It might get messy with the escapes, etc.  
That said, I'm sure you all regex wizards know what you're doing :) Fwiw though, I'd be willing to rewrite it without usage of `re` module.

LecrisUT:
> I like the functionality, but this might be too much regex magic from my PoV. It might get messy with the escapes, etc.

That was an initial concern for me as well, but check the implementation of `fmf.utils.split_pattern_repl`. There there is no escaping (other than that coming from yaml, but that's why there is is the documentation suggestion of using `'` to reduce escaping).

> Fwiw though, I'd be willing to rewrite it without usage of `re` module

Problem would be with regex groups (named/unnamed). I think that's an important feature to support for this.

martinhoyer:
> Problem would be with regex groups (named/unnamed). I think that's an important feature to support for this.

@LecrisUT No idea what what those are. Any chance you could enlighten me by pointing to how are they being used here, pretty please?

LecrisUT:
> > Problem would be with regex groups (named/unnamed). I think that's an important feature to support for this.
> 
> @LecrisUT No idea what what those are. Any chance you could enlighten me by pointing to how are they being used here, pretty please?

[An example](https://regex101.com/r/GhMHPw/1) is worth a hundred words :P. See the `()` in the search string and the `$1` in the replace.

martinhoyer:
> > > Problem would be with regex groups (named/unnamed). I think that's an important feature to support for this.
> > 
> > 
> > @LecrisUT No idea what what those are. Any chance you could enlighten me by pointing to how are they being used here, pretty please?
> 
> [An example](https://regex101.com/r/GhMHPw/1) is worth a hundred words :P. See the `()` in the search string and the `$1` in the replace.

Thanks. I still don't get why it's needed, but guessing it is important when you want the users to be able to use regex. I was thinking more about allowing just some "simple" substitutions. :)

psss:
Fixed a few typos and proposing some minor adjustments in 3a8ee40.

fmf/utils.py
```diff
@@ -894,3 +894,36 @@ def validate_data(data, schema, schema_store=None):
...
+    Delimiter set by the first character of the input and this character cannot be used
+    in the PATTERN or REPL text. Escaping is not supported.
+    """
+
+    try:
+        delimiter = source[0]
+        if source[-1] != delimiter:
+            raise FormatError("'{0}' has to end with '{1}'.".format(source, delimiter))
+    except IndexError:
+        raise FormatError("'{0}' has to start and end with '/'.".format(source))
```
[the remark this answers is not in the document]
psss: This sounds good to me.

fmf/base.py
```diff
@@ -190,6 +190,49 @@ def _merge_plus(self, data, key, value, prepend=False):
...
+            value = [value]
+        if isinstance(data[key], list):
+            data[key] = [item for item in data[key] if not lazy_any_search(item, value)]
+        elif isinstance(data[key], str):
+            if lazy_any_search(data[key], value):
+                data[key] = ''
+        elif isinstance(data[key], dict):
+            for k in list(data[key].keys()):
+                if lazy_any_search(k, value):
+                    data[key].pop(k, None)
```
psss, line 230:
```suggestion
                    data[key].pop(k)
```

fmf/base.py
```diff
@@ -190,6 +190,49 @@ def _merge_plus(self, data, key, value, prepend=False):
...
+                if re.search(p, str(item)):
+                    return True
+            return False
+        if isinstance(value, str):
+            value = [value]
+        if isinstance(data[key], list):
+            data[key] = [item for item in data[key] if not lazy_any_search(item, value)]
+        elif isinstance(data[key], str):
+            if lazy_any_search(data[key], value):
+                data[key] = ''
```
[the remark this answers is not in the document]
psss: Yeah, documenting this explicitly won't hurt.

fmf/utils.py
```diff
@@ -894,3 +894,37 @@ def validate_data(data, schema, schema_store=None):
...
+
+    Delimiter set by the first character of the input and this character
+    cannot be used in the PATTERN or REPLACEMENT text. Escaping is not
+    supported.
+    """
+
+    try:
+        delimiter = source[0]
+        if source[-1] != delimiter:
+            raise FormatError("'{0}' has to end with '{1}'.".format(source, delimiter))
```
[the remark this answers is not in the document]
lukaszachy: Because rest of fmf is using format.. Time to move forward?

lukaszachy:
Does someone explain how can fedora-40 and c9s fail on `  TypeError: adjust() got an unexpected keyword argument 'additional_rules'` but rawhide and f-39 pass?

LecrisUT:
Could be a race condition. Packit uses the merge commit and not the source commit iirc. But the srpm should be the same for all builds so not quite sure

lukaszachy:
> Could be a race condition. Packit uses the merge commit and not the source commit iirc. But the srpm should be the same for all builds so not quite sure

Ah, I guess you are right. Job ran 4h ago  and merge (introducing additional_rules) happened also around that time.

lukaszachy:
@psss I used 'repl' because ` re.sub(pattern, repl, string, count=0, flags=0)` is used in Python itself. But OK to keep longer name.

fmf/base.py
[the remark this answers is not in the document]
lukaszachy: lets see. 9856cba

docs/features.rst
```diff
@@ -121,6 +121,40 @@ attribute, matching regular expressions are replaced by an empty
...
+      - '/python2-/python3-/'
+
+__ https://docs.python.org/3/library/re.html#re.sub
+
+Remove parent value only if it matches regular expression is done
+using  the ``-~`` suffix. If value matches any of provided
+`regular expressions`__ it is removed. In the case of parent value
+being a list the item is removed from this list. If the parent value
+is a string value is set to the empty string. If parent value is
+a dictionary, the key is removed. These regular expressions can
```
LecrisUT, line 149:
```suggestion
`regular expressions`__ it is removed. If the parent value
is a list, the matching item is removed from this list. If the parent value
is a string, the value is set to an empty string. If the parent value is
a dictionary, the matching key is removed. These regular expressions can
```

Small tweaks.

I just noticed that basically the `-` operator for string does the same thing :thinking:

LecrisUT (approved):

psss (approved):
Thanks for all the adjustments!

PR #235, opened 2024-05-03
Extend the `filter()` documentation slightly

The-Mule:
People tend to try using parentheses in their filter expressions for obvious reasons. But it is not possible. I feel like this should be mentioned in the documentation. In fact, with DNF you never need parentheses to express any expression and I think this fact also needs to be noted there.

### DIFF
--- a/fmf/utils.py
+++ b/fmf/utils.py
@@ -215,12 +215,20 @@ def filter(filter, data, sensitive=True, regexp=False, name=None):
     """
     Apply advanced filter on the provided data dictionary
 
-    Filter supports disjunctive normal form with '|' used for OR, '&'
-    for AND and '-' for negation. Individual values are prefixed with
-    'value:', leading/trailing white-space is stripped. For example::
+    Filter supports disjunctive normal form (DNF) with ``|`` used for
+    OR, ``&`` for AND and ``-`` for negation. Individual values are
+    prefixed with ``key:``, leading/trailing white-space is stripped.
+    For example::
 
         tag: Tier1 | tag: Tier2 | tag: Tier3
         category: Sanity, Security & tag: -destructive
+
+    Grouping boolean expressions using parentheses is not supported but
+    the correct DNF order of precedence of operators is respected
+    (negation first, AND next and then OR). For example, to express
+    ``(tag: A | tag: B) & tag: C`` use the following filter::
+
+        tag: A & tag: C | tag: B & tag: C
 
     Use the back slash character ``\\`` to escape the boolean
     operators if you need to specify them as part of the filter


fmf/utils.py
```diff
@@ -321,7 +331,7 @@ def check_clause(clause):
 
     # At least one clause must be true
     return any([check_clause(clause)
-                for clause in re.split(r"\s*\|\s*", filter)])
+                for clause in re.split(r"\s*\|\s*(?=[^|]*:)", filter)])
```
lukaszachy, line 334: @The-Mule Could you explain why is this change necessary, this isn't just PR for docs then.

lukaszachy (changes requested):

fmf/utils.py
The-Mule: Uh oh, good catch, this slipped in by mistake, sorry - this was bit related to https://github.com/teemtee/fmf/pull/219#issuecomment-2091219335, I'll remove it.

The-Mule: Fixed.

fmf/utils.py
```diff
@@ -214,15 +214,23 @@ def filter(filter, data, sensitive=True, regexp=False):
     """
     Return true if provided filter matches given dictionary of values
 
-    Filter supports disjunctive normal form with '|' used for OR, '&'
-    for AND and '-' for negation. Individual values are prefixed with
-    'value:', leading/trailing white-space is stripped. For example::
+    Filter supports disjunctive normal (DNF) form with '|' used for OR,
```
psss, line 217:
```suggestion
    Filter supports disjunctive normal form (DNF) with '|' used for OR,
```

Let's put the abbreviation after the full term?

fmf/utils.py
```diff
@@ -214,15 +214,23 @@ def filter(filter, data, sensitive=True, regexp=False):
...
+    '&' for AND and '-' for negation. Individual values are prefixed
+    with 'value:', leading/trailing white-space is stripped. For
+    example:
 
         tag: Tier1 | tag: Tier2 | tag: Tier3
         category: Sanity, Security & tag: -destructive
 
+    Even though parentheses are only allowed in keys or values they are
```
psss, line 225: Parenthesis are supported in keys or values? Do you mean like regular characters? 'cause, I think currently there's not support for grouping at all.

fmf/utils.py
```diff
@@ -214,15 +214,23 @@ def filter(filter, data, sensitive=True, regexp=False):
...
+    not needed for DNF as the correct order of precedence of operators
+    is respected (negation first, AND next and then OR). For example,
+    to express '(tag: A | tag: B) & tag: C' use the following filter:
```
psss, line 228:
```suggestion
    to express '(tag: A | tag: B) & tag: C' use the following filter::
```

Double colons are need for the online documentation to be properly rendered:
* https://fmf--235.org.readthedocs.build/en/235/modules.html#fmf.filter

fmf/utils.py
```diff
@@ -236,6 +244,8 @@ def filter(filter, data, sensitive=True, regexp=False):
     filter is not found in the data dictionary. Set option 'sensitive'
     to False to enable case-insensitive matching. If 'regexp' option is
     True, regular expressions can be used in the filter values as well.
+    However, using either '|' or '&' in the regular expression is not
+    allowed.
```
psss, line 248: So I guess, the `|` could/should be covered by #220 which should be included in the `1.4` release as well. Shall we extend the pull request to support `&` escaping as well?

psss (changes requested):
Thanks for improving the docs. Added a couple of questions and suggestions.

lukaszachy:
Depends on #220, we have to adjust docs once that is merged.

The-Mule:
Thanks for the fix, Petre, let me also rebase it now.

psss:
Thanks for the rebase, going to squash the changes now.

lukaszachy (approved):
One nitpick, but no point in blocking on that

PR #238, opened 2024-05-21
Simplify spec file

LecrisUT:
Depends-on #202


LecrisUT:
Nevermind, #202 does not build without it

Issue #246, opened 2024-06-17
py3.13 - `fmf ls --condition 'float(release) >= 7'` broken


lukaszachy:
Used in `tests/unit/test_cli.py` - `test_conditions`. Can be reproduced easily:
```
$ cd examples/conditions
$ fmf ls --condition 'flaot(release) >= 7'
```

On python3.12 and older it prints 3 tests. On python3.13 the output is empty.

Blocks fedora rawhide - https://bugzilla.redhat.com/show_bug.cgi?id=2279994

lukaszachy:
And the reason of the change is https://docs.python.org/3.13/whatsnew/3.13.html#whatsnew313-locals-semantics

PR #247, opened 2024-06-17
New way to populate locals() for filter

lukaszachy:
Details in https://docs.python.org/3.13/whatsnew/3.13.html#whatsnew313-locals-semantics

Fix: #246

### DIFF
--- a/fmf/utils.py
+++ b/fmf/utils.py
@@ -202,9 +202,11 @@ def evaluate(expression, data, _node=None):
     Expects data dictionary which will be used to populate local
     namespace. Used to provide flexible conditions for filtering.
     """
-    locals().update(data)
+    # Since python3.13 and https://peps.python.org/pep-0667/
+    _locals = copy.deepcopy(locals())
+    _locals.update(data)
     try:
-        return eval(expression)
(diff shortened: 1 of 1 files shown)


fmf/utils.py
```diff
@@ -202,9 +202,11 @@ def evaluate(expression, data, _node=None):
     Expects data dictionary which will be used to populate local
     namespace. Used to provide flexible conditions for filtering.
     """
-    locals().update(data)
+    # Since python3.13 and https://peps.python.org/pep-0667/
+    _locals = copy.deepcopy(locals())
```
happz, line 206: I don’t `deepcopy()` is needed. `locals()` should regurn an independent snapshot of local namespace, i.e. we would be making (and updating) a copy of dictionary no other part of code can ever see. Seems superfluous.

happz: In other words, I agree we need to store `locals()` somewhere if we need to both update it and pass it to a function call,  ut IIUIC, we don’t need to work with a copy of said dictionary.

lukaszachy:
Drop in favor of #248

PR #240, opened 2024-05-21
Add ruff

LecrisUT:
Depends-on #202 for the PEP621 of course

Copied all of the configuration from tmt and I have disabled a bunch of tests just so that we can bring this in first and then do each one in a separate PR

PR #253, opened 2024-08-27
Use the `referencing` library where available

martinhoyer:
Use the `referencing` library where available instead of deprecated `jsonschema.RefResolver`.

Addressing https://github.com/teemtee/fmf/issues/194 .

Getting rid of deprecation warnings.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,7 +36,7 @@ share/python-wheels/
 .installed.cfg
 *.egg
 MANIFEST
-tmt.1
+fmf.1
 
 # Testing
 .mypy_cache
--- a/.gitignore
+++ b/.gitignore
@@ -9,4 +9,77 @@ dist
 docs/_build
 docs/spec
 docs/stories
-__pycache__
+
+# Python
+
+# Byte-compiled / optimized / DLL files
+__pycache__/
+*.py[cod]
+*$py.class
+
+# Distribution / packaging
+.Python
+build/
+develop-eggs/
+dist/
+downloads/
+eggs/
+.eggs/
+lib/
+lib64/
+parts/
+sdist/
+var/
+wheels/
+share/python-wheels/
+*.egg-info/
+.installed.cfg
+*.egg
+MANIFEST
+tmt.1
+
+# Testing
+.mypy_cache
+.pytest_cache
+*.tgz
+# Created by pytest-html reporting plugin
+/assets/style.css
+/report.html
+
+# Virtual environment
+.env
+.venv
+env/
+venv/
+ENV/
+env.bak/
+venv.bak/
+
+# Jetbrains
+.idea/
+
+# Vim
+
+# Swap
+[._]*.s[a-v][a-z]
+[._]*.sw[a-p]
+[._]s[a-rt-v][a-z]
+[._]ss[a-gi-z]
+[._]sw[a-p]
+
+# Session
+Session.vim
+Sessionx.vim
+
+# Temporary
+.netrwhist
+*~
+
+# Auto-generated tag files
+tags
+
+# Persistent undo
+[._]*.un~
+
+# Visual Studio Code
+.vscode
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -9,7 +9,6 @@ repos:
           - --recursive
           - --in-place
           - --aggressive
-          - --aggressive
           - --hang-closing
           - --max-line-length=99
 
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -9,6 +9,7 @@ repos:
           - --recursive
           - --in-place
           - --aggressive
+          - --aggressive
           - --hang-closing
           - --max-line-length=99
 
--- a/fmf/utils.py
+++ b/fmf/utils.py
@@ -10,7 +10,7 @@ import sys
 import time
 import warnings
 from io import StringIO
-from typing import Any, List, NamedTuple
+from typing import Any, NamedTuple, Optional
 
 import jsonschema
 from filelock import FileLock, Timeout
@@ -18,6 +18,7 @@ from ruamel.yaml import YAML, scalarstring
 from ruamel.yaml.comments import CommentedMap
 
 import fmf.base
+from fmf._compat.jsonschema import get_validator
 
 # ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 #  Constants
@@ -419,7 +420,7 @@ class Logging:
     _level = LOG_WARN
 
     # Already initialized loggers by their name
-    _loggers = dict()
+    _loggers: dict[str, logging.Logger] = dict()
 
     def __init__(self, name='fmf'):
         # Use existing logger if already initialized
@@ -912,34 +913,22 @@ def dict_to_yaml(data, width=None, sort=False):
 # ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 class JsonSchemaValidationResult(NamedTuple):
-    """ Represents JSON Schema validation result """
-
+    """Represents JSON Schema validation result."""
     result: bool
-    errors: List[Any]
+    errors: list[Any]
 
 
-def validate_data(data, schema, schema_store=None):
+def validate_data(
+        data: Any,
+        schema: Any,
+        schema_store: Optional[dict[str, Any]] = None
+        ) -> JsonSchemaValidationResult:
+    """Validate data against a JSON Schema.
+
+    Validates the given data using the specified JSON Schema and optional
+    schema references.
     """
-    Validate data with given JSON Schema and schema references.
-
-    schema_store is a dict of schema references and their content.
-
-    Return a named tuple utils.JsonSchemaValidationResult
-    with the following two items:
-
-        result ... boolean representing the validation result
-        errors ... A list of validation errors
-
-    Raises utils.JsonSchemaError if the supplied schema was invalid.
-    """
-    schema_store = schema_store or {}
-    try:
-        resolver = jsonschema.RefResolver.from_schema(
-            schema, store=schema_store)
-    except AttributeError as error:
-        raise JsonSchemaError(f'Provided schema cannot be loaded: {error}')
-
-    validator = jsonschema.Draft4Validator(schema, resolver=resolver)
+    validator = get_validator(schema, schema_store)
 
     try:
         validator.validate(data)
@@ -955,6 +944,7 @@ def validate_data(data, schema, schema_store=None):
             jsonschema.exceptions.RefResolutionError,
             jsonschema.exceptions.UnknownType
             ) as error:
+        from fmf.utils import JsonSchemaError
         raise JsonSchemaError(f'Errors found in provided schema: {error}')
 
 
--- a/fmf/utils.py
+++ b/fmf/utils.py
@@ -913,7 +913,7 @@ def dict_to_yaml(data, width=None, sort=False):
 # ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 class JsonSchemaValidationResult(NamedTuple):
-    """Represents JSON Schema validation result."""
+    """ Represents JSON Schema validation result """
     result: bool
     errors: list[Any]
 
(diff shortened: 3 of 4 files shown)


fmf.spec
```diff
@@ -34,6 +34,7 @@ BuildRequires: python%{python3_pkgversion}-jsonschema
 BuildRequires: git-core
 %{?python_provide:%python_provide python%{python3_pkgversion}-%{name}}
 Requires:       git-core
+Recommends:     python%{python3_pkgversion}-referencing
```
LecrisUT, line 37: I don't think we need it. The dependency is part of `jsonschema` as early as 4.18 [^1] and for rhel case we can just assume it does not have the interface.

[^1]: https://github.com/python-jsonschema/jsonschema/commit/69f3899c0770472f19898d80aea2b7a51fd5ec6b

fmf/_compat_jsonschema.py
```diff
@@ -0,0 +1,91 @@
...
+def create_validator(
+        schema: Dict[str, Any],
+        schema_store: Optional[Dict[str, Any]] = None
+        ) -> Union[jsonschema.validators.Validator, jsonschema.validators.Draft4Validator]:
+    """Create a validator instance based on available jsonschema version."""
+    schema_store = schema_store or {}
+
+    if MODERN_JSONSCHEMA:
+        # Modern approach with referencing
+        registry: Registry = Registry().with_resources([
```
LecrisUT, line 34:
```suggestion
        registry = Registry().with_resources([
```

Should not be necessary

fmf/_compat_jsonschema.py
```diff
@@ -0,0 +1,91 @@
...
+
+import jsonschema
+
+try:
+    from jsonschema import validators
+    from referencing import Registry, create
+    from referencing.jsonschema import DRAFT4
+    MODERN_JSONSCHEMA = True
+except ImportError:
+    MODERN_JSONSCHEMA = False
```
LecrisUT, line 16: Could we make the test be on the `importlib` package version instead? It is less clear what `MODERN_JSONSCHEMA` is referring to otherwise. As for the `create_validator`, what about adding the imports in there?

fmf/_compat_jsonschema.py
```diff
@@ -0,0 +1,91 @@
+"""Compatibility layer for jsonschema validation."""
+
+from __future__ import annotations
+
+import warnings
+from typing import Any, Dict, List, NamedTuple, Optional, Union
```
LecrisUT, line 6: Let's not mix `__future__.annotation` and `typing.Dict` for now. We should take a hammer at it eventually though.

fmf/_compat_jsonschema.py
```diff
@@ -0,0 +1,91 @@
...
+    from referencing.jsonschema import DRAFT4
+    MODERN_JSONSCHEMA = True
+except ImportError:
+    MODERN_JSONSCHEMA = False
+
+
+class JsonSchemaValidationResult(NamedTuple):
+    """Represents JSON Schema validation result."""
+    result: bool
+    errors: List[Any]
```
LecrisUT, line 22: I don't think these should be in a `_compat` package. Maybe `utils.jsonschema` instead. For this PR though, we can keep them in the original place.

fmf/_compat_jsonschema.py
```diff
@@ -0,0 +1,91 @@
...
+        registry: Registry = Registry().with_resources([
+            (uri, create({'$schema': 'http://json-schema.org/draft-04/schema#', **contents}))
+            for uri, contents in schema_store.items()
+            ])
+        cls = validators.create(meta_schema=DRAFT4, validators={})
+        return cls(schema=schema, registry=registry)
+    else:
+        # Legacy approach with RefResolver
+        with warnings.catch_warnings():
+            warnings.filterwarnings('ignore', category=DeprecationWarning)
```
LecrisUT, line 43: Let's try without the `catch_warnings`. There should not be the DeprecationWarning being triggered if the `MODERN_JSONSCHEMA` is working appropriately, and if it does, we should investigate why it does.

fmf/_compat_jsonschema.py
```diff
@@ -0,0 +1,91 @@
...
+    """Represents JSON Schema validation result."""
+    result: bool
+    errors: List[Any]
+
+
+def create_validator(
+        schema: Dict[str, Any],
+        schema_store: Optional[Dict[str, Any]] = None
+        ) -> Union[jsonschema.validators.Validator, jsonschema.validators.Draft4Validator]:
+    """Create a validator instance based on available jsonschema version."""
```
LecrisUT, line 29: I am more used to `create` naming convention being a static/classmethod. Just a thought if it should be `get` or `create` or something else.

fmf/utils.py
```diff
@@ -10,14 +10,17 @@
...
-import jsonschema
 from filelock import FileLock, Timeout
 from ruamel.yaml import YAML, scalarstring
 from ruamel.yaml.comments import CommentedMap
 
 import fmf.base
+# F401 utils is entry point, while _compat is being used
+# isort with autopep8 made it to two lines. Ruff needed.
+from fmf._compat_jsonschema import JsonSchemaValidationResult  # noqa: F401
+from fmf._compat_jsonschema import validate_data  # noqa: F401
```
LecrisUT, line 23: Why didn't `fmf._compat.jsonschema` work?

fmf/utils.py
```diff
@@ -419,7 +422,7 @@ class Logging:
     _level = LOG_WARN
 
     # Already initialized loggers by their name
-    _loggers = dict()
+    _loggers: dict[str, logging.Logger] = dict()
```
LecrisUT, line 425: This doesn't have `__future__.annotations` here so it should be using `Dict` instead here. But also let's deffer the change.

LecrisUT (changes requested):

fmf.spec
martinhoyer: oh cool, thanks. Removing.

fmf/utils.py
martinhoyer: I think `dict` is already in Python 3.9, no?

happz: Correct, Python 3.9 allows builtin types as annotations.

happz:
So many distinct changes squashed into a single PR :(

martinhoyer:
> So many distinct changes squashed into a single PR :(

Well, I couldn't run pre-commit without it and didn't want to make workarounds.

martinhoyer:
...but fwiw, for there is a pr for the last commit https://github.com/teemtee/fmf/pull/258

Context for the autopep8 is that the previous hook repo was unmaintained and stopped being updated (doesn't work at all on py3.13). Replaced it with what was recommended as replacement and with `--strict`, the `is_instance` changes were needed to pass that check.

fmf/_compat_jsonschema.py
martinhoyer: How about now, without the flag?

fmf/_compat_jsonschema.py
martinhoyer: moved back, thanks

fmf/_compat_jsonschema.py
martinhoyer: True. Removed.

fmf/_compat_jsonschema.py
martinhoyer: renamed

fmf/utils.py
martinhoyer: hmm, I don't remember. Any suggestion about how the structure should look like are welcome.

fmf/utils.py
LecrisUT: Ah right, must have confused it with `Union` compatibility issue

fmf/utils.py
LecrisUT: The design I saw was have a `_compat` package and under that the same structure as the most recent interface that will eventually replace it, e.g. `_compat.collections.abc` for `Itterator`, `_compat.warnings` for `deprecated` etc. That way the eventual change is to simply drop the `_compat` from the `from import`.

The part that I didn't see used was something like `from _compat import warnings` that would cover both compatibilities like `warnings.deprecated` and normal functions like `warnings.warn`. It seems sensible to me if you want to eventually use the fully-qualified naming, but not sure if there are pitfalls that I'm overlooking.

fmf/utils.py
```diff
@@ -955,6 +957,7 @@ def validate_data(data, schema, schema_store=None):
             jsonschema.exceptions.RefResolutionError,
             jsonschema.exceptions.UnknownType
             ) as error:
+        from fmf.utils import JsonSchemaError
```
LecrisUT, line 960: Unneeded? We are in `utils.py`

fmf/utils.py
```diff
@@ -912,34 +915,33 @@ def dict_to_yaml(data, width=None, sort=False):
...
+        data: The data to validate
+        schema: The JSON Schema to validate against
+        schema_store: Optional dict of schema references and their content
 
-        result ... boolean representing the validation result
-        errors ... A list of validation errors
+    Returns:
+        JsonSchemaValidationResult with:
+            result: boolean representing the validation result
+            errors: A list of validation errors
```
LecrisUT, line 939: What docstyle has been adopted?

fmf/_compat_jsonschema.py
```diff
@@ -0,0 +1,51 @@
...
+
+def get_validator(
+        schema: Any,
+        schema_store: Optional[dict[str, Any]] = None
+        ) -> Draft4Validator:
+    """Create a validator instance based on available jsonschema version."""
+    # Validate schema is a dict
+    if not isinstance(schema, dict):
+        from fmf.utils import JsonSchemaError
+        raise JsonSchemaError(f'Invalid schema type: {type(schema)}. Schema must be a dictionary.')
```
LecrisUT, line 17: Probably this check can be removed when `mypy` is added. Should we add a todo note for that?

fmf/_compat_jsonschema.py
LecrisUT: 👍

psss:
> > So many distinct changes squashed into a single PR :(
> 
> Well, I couldn't run pre-commit without it and didn't want to make workarounds.

Let's address at least the `pre-commit` changes separately in #259?

fmf/utils.py
martinhoyer: hah, you have to ask my IDE, I'm just pressing TAB :D  
Joking aside, I do rely on formatters and linters for this stuff. :/

martinhoyer: @LecrisUT maybe we can move https://github.com/teemtee/fmf/pull/240 forward?

fmf/utils.py
LecrisUT: You forgot this one in the recent rebase?

fmf/utils.py
LecrisUT: Lgtm, don't think we can follow the other comoat designs here, but at least the change is localized so it helps to be in this form 👍

LecrisUT (approved):

fmf/utils.py
martinhoyer: Huh, I specifically checked for this. Must be going insane. Thanks mate.

.gitignore
```diff
@@ -9,4 +9,77 @@ dist
...
+parts/
+sdist/
+var/
+wheels/
+share/python-wheels/
+*.egg-info/
+.installed.cfg
+*.egg
+MANIFEST
+tmt.1
```
psss, line 39: Irrelevant copy & paste here? Should be `fmf.1` instead?

.pre-commit-config.yaml
```diff
@@ -9,7 +9,6 @@ repos:
           - --recursive
           - --in-place
           - --aggressive
-          - --aggressive
```
psss, line 12: Why this change?

fmf/utils.py
```diff
@@ -912,34 +913,22 @@ def dict_to_yaml(data, width=None, sort=False):
 # ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 class JsonSchemaValidationResult(NamedTuple):
-    """ Represents JSON Schema validation result """
-
+    """Represents JSON Schema validation result."""
```
psss, line 916: Can we keep these as they are until there's a concensus on the formatting?

psss (changes requested):
Thanks for looking into this! Looks good, just a few minor comments.

.gitignore
martinhoyer: thanks

.pre-commit-config.yaml
martinhoyer: it's back. Thanks.

fmf/utils.py
martinhoyer: reverted in fixup.

fmf/utils.py
psss: > @LecrisUT maybe we can move #240 forward?

Yeah, let's enable it in `1.6`.

psss (approved):

PR #254, opened 2024-09-03
Handle missing attribute for merge/minus operation

lukaszachy:
Operations `~`, `-~` and `-` should do nothing if the key is not present. Otherwise they are difficult to use with shared adjust snippets as keys might not always be present.

### DIFF
--- a/fmf/base.py
+++ b/fmf/base.py
@@ -173,7 +173,7 @@ class Tree:
     def _merge_plus(self, data, key, value, prepend=False):
         """ Handle extending attributes using the '+' suffix """
 
-        # Nothing to do if key not in parent
+        # Set the value if key is not present
         if key not in data:
             data[key] = value
             return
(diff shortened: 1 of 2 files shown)


lukaszachy:
Added to the milestone as I intend to use this with 'adjust-tests' feature of tmt and it is very easy to hit KeyError when some test doesn't have both require/recommend set...

LecrisUT:
Can you comment with an example fmf file for this?

lukaszachy:
For example conflicting mariadb vs mysql server but test can do both... Recommend package can be skipped, but if for some reason you change the order -> failure during prepare.
```
/1:
  test: echo
  require:
   - mysql
/2:
  test: echo
  recommend:
   - mariadb
   - mysql
```

So if I was to use this from other plan (and I don't have possiblity to "fix" test metadata (eg. not my repo):

```
discover:
 how: fmf
 adjust-tests:
 - require-: [mysql]
 - recommend-: [mysql]
 ```

And now I can be sure that mysql isn't require by any test..

lukaszachy:
@happz has good points that this might hide typo in the key names, making it harder to spot (e.g. why there still is 'foo' in require why I clearly have 'requre-:[foo]' in data...)

psss:
So, a bit cleaner (perhaps) approach discussed on the chat would be to use `require: []` at the top of the test repository to make sure that the key is always defined. But as @lukaszachy mentioned above, there might be a scenario when user does not have write access to the test repository. And for that use case the unconditional removal could be really useful.

psss (approved):
Looks good & makes sense. Thanks!

PR #255, opened 2024-10-31
Allow dashes in context dimension names

psss:
It's quite common to name multi-word keys with dashes instead of underscores. Let's allow dashes in the context dimension names as well.


lukaszachy (approved):

lukaszachy:
@psss any objections to squeeze this into 1.6.1 ?

PR #256, opened 2024-11-07
Add support for +* operation

kkaarreell:


kkaarreell:
it is questionable if we really need a new operator. Can't we simply stick to `+`?

kkaarreell:
```
$ cat main.fmf 
discover:
    - how: fmf
      url: https://github.com/project1
    - how: fmf
      url: https://github.com/project2

/basic:
    summary: Quick set of basic functionality tests
    discover+*:
        filter: "tier:1"

/features:
    summary: Detailed tests for individual features
    discover+*:
        filter: "tier:2"
ksrot@localhost fmf $ fmf show
/basic
discover: [{'filter': 'tier:1', 'how': 'fmf', 'url': 'https://github.com/project1'},
 {'filter': 'tier:1', 'how': 'fmf', 'url': 'https://github.com/project2'}]
summary: Quick set of basic functionality tests

/features
discover: [{'filter': 'tier:2', 'how': 'fmf', 'url': 'https://github.com/project1'},
 {'filter': 'tier:2', 'how': 'fmf', 'url': 'https://github.com/project2'}]
summary: Detailed tests for individual features
```

kkaarreell:
So, after reading through the original issue it seems there are more people favoring using the `+` operator.
With that in mind, I have pushed another commit which modified the code so  that only `+` is used.

I like this option more (if you want to compare both variants see the 1st commit only).

Moreover, now the following use case seems to be handled as well:
```
discover:
    how: fmf
    prune: True
    adjust-tests:
        - duration+: '*3'
          when: hw == slow

/path:
  discover+:
    - name: upstream
      url: https://some.url
    - name: downstream
      url: https://other.url
```
```
$ fmf show
/path
discover: [{'adjust-tests': [{'duration+': '*3', 'when': 'hw == slow'}],
  'how': 'fmf',
  'name': 'upstream',
  'prune': True,
  'url': 'https://some.url'},
 {'adjust-tests': [{'duration+': '*3', 'when': 'hw == slow'}],
  'how': 'fmf',
  'name': 'downstream',
  'prune': True,
  'url': 'https://other.url'}]
```

fmf/base.py
```diff
@@ -172,10 +172,29 @@ def _initialize(self, path):
...
+                    "MergeError: Item '{0}' in {1} must be a dictionary.".format(
+                        list_item, self.name)
+                self._merge_special(list_item, value)
+            return
+        if isinstance(data[key], dict) and isinstance(value, list):
+            for list_item in value:
+                if not isinstance(list_item, dict):
+                    "MergeError: Item '{0}' in {1} must be a dictionary.".format(
+                        list_item, self.name)
+                self._merge_special(list_item, data[key])
```
kkaarreell, line 194: Just wondering, is this safe? I am modifying item from the list I am iterating through. But I am not adding any new items to the list so I hope this is fine. Or should I do a deepcopy instead?

happz: You should be safe as long as you are not changing the list itself, i.e. adding to it, removing from it, or replacing items with other objects. Changing an item itself, if it's a mutable value, should not cause any trouble.

kkaarreell: So I guess we should be fine.

kkaarreell, line 194: Btw, what's happening with those original `something+` attributes? I mean, we do the update of the respective `something` attribute but what happens with `something+`. Does `fmf` care about it? Does it matter if we modify it? I am asking because this code is in fact modifying the `something+` element as well.

KwisatzHaderach (approved):

kkaarreell:
@psss Hi, I do not plan any further updates unless requested. Please review.

kkaarreell:
@psss Hi, could you please check failing tests? It doesn't seem related, rather like a change in Fedora distro that is not reflected in tests. Thank you.

kkaarreell:
Tests [failures](https://artifacts.dev.testing-farm.io/52c88d04-a600-48b2-972a-b48da24d8645/) doesn't seem to be related. Same [failures were present](https://artifacts.dev.testing-farm.io/ba15837f-082c-48f6-9fdd-77b6d9c924d8/) even when I have reverted my commits.

fmf/base.py
psss: The `_merge_special()` method should take target data as the first argument and the second one as the source of the modification. Here, I believe, we need to create several copies of the parent dictionary and update them with the respective list items. Should be addressed in e2b9c19. @kkaarreell, please, have a look if everything's ok.

kkaarreell: I have done in the other way around. I have modified list items with dict instead and later replaced the original dict with the newly updated list. But I think that the `deepcopy` approach is better.

PR #258, opened 2024-11-26
Fix integration test plan

martinhoyer:
I think this was making CI 'not green'.

### DIFF
--- a/plans/integration.fmf
+++ b/plans/integration.fmf
@@ -9,17 +9,5 @@ prepare:
   - how: install
     package:
       - jq
-      - python3-pip
+      - yq
       - tmt-all
-  - how: shell
-    script:
-      # Try to install into .local first, if that fails, try without --user.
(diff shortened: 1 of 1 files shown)


martinhoyer:
hmm, looks like that's not it. Digging deeper.

martinhoyer:
Nevermind, apparently we need to wait for tmt 1.39.0 to be in stable for these tests to work.

LecrisUT:
`yq` part though should be simplified like that though. As for tmt 1.39.0, that should fix itself when the design of the copr project changes and are unified in `latest` and `stable`. Just for curiosity can you add the following change in `.packit.yaml` to see if it would work as intended?
```patch
   - job: copr_build
     trigger: pull_request
+    additional_repos: [ "copr://teemtee/tmt" ]
     targets:
     - fedora-all
     - epel-9
```

PR #260, opened 2024-12-02
Keep order of the operation when merging values

### LINKED ISSUE #232: Inconsistent order of operations
Example:
```yaml
my_var:
  - one
  - two

/overwrite:
  my_var-: [ two, three ]
  my_var+: [ three, four ]
```

This one gives `my_var: one and four` which would be inconsistent based on the order of definition. This might not be that crucial, but at least it should be documented that the keys are ordered before being parsed by fmf, thus the order is ...

Are there tests at least covering the operation order based on inheritance (inherited operations occur first, e.g. if `my_var-:` was one level below, `three` should be in the list) and adjust (after base variable operations)?

psss:
When inheriting, do not sort the keys but apply the merging operation in the original order provided by the user.

Fix #232.

### DIFF
--- a/fmf/base.py
+++ b/fmf/base.py
@@ -294,7 +294,7 @@ class Tree:
 
     def _merge_special(self, data, source):
         """ Merge source dict into data, handle special suffixes """
-        for key, value in sorted(source.items()):
+        for key, value in source.items():
             # Handle special attribute merging
             if key.endswith('+'):
                 self._merge_plus(data, key.rstrip('+'), value)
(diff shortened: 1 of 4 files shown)


seberm (approved):

psss:
Integration test `/tests/core/order` revealed that we've changed order of the nodes when `climb()` or `prune()` is called. Should be fixed by d1511f1.

fmf/base.py
```diff
@@ -671,7 +671,7 @@ def climb(self, whole=False):
         """ Climb through the tree (iterate leaf/all nodes) """
         if whole or self.select:
             yield self
-        for name, child in self.children.items():
+        for name, child in sorted(self.children.items()):
```
kkaarreell, line 674: @psss So why is `sorted` here?

martinhoyer (approved):
I'm not super familiar with the relevant code, but make sense to me. My only question would be the same as Karel's

kkaarreell (approved):

PR #262, opened 2024-12-02
Release fmf-1.5.0

psss:
### Follow the steps below to create a new major or minor release:

- [x] Update README.rst with new contributors since the last release
- [x] Review the release notes in releases.rst, update as needed
- [x] Add a Release x.y.z commit, empty if needed: git commit --allow-empty -m "Release x.y.z"
- [x] Create a pull request with the commit, ensure tests pass, merge it

### Create a new github release based on the tag above

- [ ] Move the fedora branch to point to the new release
- [ ] Tag the commit with x.y.z, push tags git push --tags
- [ ] Create a new github release based on the tag above
- [ ] Mention the most important changes in the name, do not include version
- [ ] Use ; as a delimiter, when multiple items are mentioned in the name
- [ ] Push the “Generate release notes” button to create the content
- [ ] Prepend the “See the release notes for the list of interesting changes.” line
- [ ] Publish the release, check Fedora pull requests, make sure tests pass and merge
- [ ] Close the corresponding release milestone

### DIFF
--- a/docs/releases.rst
+++ b/docs/releases.rst
@@ -12,6 +12,51 @@ The fmf :ref:`trees` can now be built from hidden files and
 directories as well. Use a simple :ref:`config` file to specify
 names which should be included in the search.
 
+The ``+`` operator now can be used for merging ``list`` of
+dictionaries with a single ``dict``. This can be for example
+useful when extending the ``discover`` step config which defines
+several phases:
+
+.. code-block:: yaml
+
+    discover:
+      - how: fmf
+        url: https://github.com/project/one
+      - how: fmf
+        url: https://github.com/project/two
+
+    /tier1:
+        summary: Run tier one tests
+        discover+:
+            filter: "tier:1"
+
+    /tier2:
+        summary: Run tier two tests
+        discover+:
+            filter: "tier:2"
+
+See the :ref:`merging<merging>` section for more details and
+examples.
+
+The ``-`` operator no more raises exception when the parent node
+does not define the key. This allows reducing values even for
+cases when user does not have write permissions for the parent
+data. For example, in order to make sure that the ``mysql``
+package is not included in the list of required or recommended
+packages you can now safely use this:
+
+.. code-block:: yaml
+
+    discover:
+        how: fmf
+        adjust-tests:
+          - require-: [mysql]
+          - recommend-: [mysql]
+
+When merging inherited values from parent, merge operations are
+now performed in the exact order in which user specified them, the
+keys are no more sorted before the merging step.
+
 
 fmf-1.4.0
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/docs/releases.rst
+++ b/docs/releases.rst
@@ -38,12 +38,12 @@ several phases:
 See the :ref:`merging<merging>` section for more details and
 examples.
 
-The ``-`` operator no more raises exception when the parent node
-does not define the key. This allows reducing values even for
-cases when user does not have write permissions for the parent
+The ``-`` operator no longer raises exception when the key is not
+defined by the parent node. This allows reducing values even for
+cases where user does not have write permissions for the parent
 data. For example, in order to make sure that the ``mysql``
 package is not included in the list of required or recommended
-packages you can now safely use this:
+packages, you can now safely use this:
 
 .. code-block:: yaml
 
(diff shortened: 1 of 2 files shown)


docs/releases.rst
```diff
@@ -4,7 +4,57 @@
...
+
+    /tier2:
+        summary: Run tier two tests
+        discover+:
+            filter: "tier:2"
+
+See the :ref:`merging<merging>` section for more details and
+examples.
+
+The ``-`` operator no more raises exception when the parent node
```
martinhoyer, line 37: "no longer raises exception when key is not defined by the parent node"?

docs/releases.rst
```diff
@@ -4,7 +4,57 @@
...
+does not define the key. This allows reducing values even for
+cases when user does not have write permissions for the parent
```
martinhoyer, line 39:
```suggestion
cases where user does not have write permissions for the parent
```

docs/releases.rst
```diff
@@ -4,7 +4,57 @@
...
+data. For example, in order to make sure that the ``mysql``
+package is not included in the list of required or recommended
+packages you can now safely use this:
```
martinhoyer, line 42:
```suggestion
packages, you can now safely use this:
```

docs/releases.rst
```diff
@@ -4,7 +4,57 @@
...
+
+.. code-block:: yaml
+
+    discover:
+        how: fmf
+        adjust-tests:
+          - require-: [mysql]
```
martinhoyer, line 49: Do we have some "standard" for single item arrays in documentation? It would be good to have the same style everywhere (not strictly related to this PR)
Reminded of https://github.com/teemtee/tmt/pull/2187 :)

docs/releases.rst
```diff
@@ -4,7 +4,57 @@
...
+          - recommend-: [mysql]
+
+When merging inherited values from parent, merge operations are
+now performed in the exact order in which user specified them, the
+keys are no more sorted before the merging step.
```
martinhoyer, line 54:
```suggestion
keys are no longer sorted before the merging step.
```

docs/releases.rst
```diff
@@ -4,7 +4,57 @@
...
+
+
+fmf-1.4.0
```
martinhoyer, line 57: I think this should stay as it was released, no?

martinhoyer (commented):
Just a few nitpicks :) Thanks!

docs/releases.rst
psss: This is just for consistency. We've switched all versions to three digits in `tmt` as well.

* teemtee/tmt/pull/3205

I don't think we need a special pull request for this.

martinhoyer: I see, there are only two releases in the docs. I assumed there are more and only the 1.4 was changed. All good then.

docs/releases.rst
psss: I don't think we have any standard. And for short lists I'd say this notation is more readable/concise.

docs/releases.rst
psss: Good wording, applied in 5111c52.

docs/releases.rst
psss: Applied in 5111c52.

docs/releases.rst
psss: Ack.

martinhoyer (approved):

LecrisUT (approved):

psss:
So, the release action [failed](https://github.com/teemtee/fmf/actions/runs/12147692786/job/33874493224) with:

```
pipx: error: argument command: invalid choice: 'hatch'
```

@LecrisUT, why `pipx` is used for running `hatch build`?

LecrisUT:
I forgot I had it like that. `pipx` is just to avoid a `pip install` command. Should have replaced it with [build-and-inspect-python-package](https://github.com/hynek/build-and-inspect-python-package) and always run it.

martinhoyer:
> I forgot I had it like that. `pipx` is just to avoid a `pip install` command. Should have replaced it with [build-and-inspect-python-package](https://github.com/hynek/build-and-inspect-python-package) and always run it.
> 
> Yes, the `pipx run` should have been there

This looks pretty good.

As for the pipx vs pip, etc...
Ultimately doesn't matter, especially since it's being run only on release. Also, depends on what's already available on ths image.

I wouldn't be afraid of just doing pip install hatch here. But yeah hynek's action looks cool.

Anyway, for 1.5.0 - manual `hatch publish`?

psss:
There might be more surprises on the way, let's try the automation until the end. I've updated the command in #268 and will re-create the release again.

PR #264, opened 2024-12-03
Explore metadata in hidden files and directories

### LINKED ISSUE #183: Support for `.dot` directories for storing plans
Seems in some cases this would be useful to users.

```
❯ find
.
./.fmf
./.fmf/version
./.foo
./.foo/bar
./.foo/bar/plan.fmf

❯ tmt plan ls
<no output>

❯ tmt
Found 0 tests.
Found 0 plans.
Found 0 stories.
```

I wonder if there is a technical limitation why this does not work.

psss:
Add support for a simple config where users can specify which hidden files and directories should be included in the search.

Fix #183.


seberm (approved):
LGTM

fmf/base.py
```diff
@@ -170,6 +174,17 @@ def _initialize(self, path):
         except ValueError:
             raise utils.FormatError("Invalid version format")
 
+        # Check for the config file
+        config_file_path = Path(self.root) / ".fmf" / "config"
```
LecrisUT, line 178: In general there is no benefit from chaining `/` operators even when parsing on windows paths. I am not sure if it's detrimental in terms of performance though.

psss: Thanks for the hint, simplified in 1efb3b5.

docs/features.rst
```diff
@@ -79,6 +79,28 @@ directive::
...
+
+.. _config:
+
+Config
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+By default, all hidden files are ignored when exploring metadata
+on the disk. If a specific file or directory should be included in
+the search, create a simple config file ``.fmf/config`` with the
+following format:
```
LecrisUT, line 91: Probably should move these sections to be together in a fmf control section where all fmf metadata can be found. Currently `.fmf/version`, `.fmf/config`, `/:` configurations, etc.

For the `explore.include` specifically, we should explicitly state that `.fmf` folder cannot be used in there. Currently we technically allow it, but it feels like it will ask for trouble in the future.

docs/releases.rst
```diff
@@ -4,7 +4,16 @@
...
 ======================
 
-fmf-1.4
+
+fmf-1.5.0
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Now it is possible to store metadata in hidden files and
+directories as well. Use a simple :ref:`config` file to specify
+names which should be included in the search.
```
LecrisUT, line 13: Store metadata part feels a bit technical for this change. Was thinking something along the lines of fmf tree can now be built with hidden node paths to reflect that `/path/.hidden/something` is the main thing that is now supported.

docs/features.rst
psss: > Probably should move these sections to be together in a fmf control section where all fmf metadata can be found. Currently `.fmf/version`, `.fmf/config`, `/:` configurations, etc.

Good, point. What about documenting `Trees` and `Config` close together? See ecd8e0f.

> For the `explore.include` specifically, we should explicitly state that `.fmf` folder cannot be used in there. Currently we technically allow it, but it feels like it will ask for trouble in the future.

Yeah, makes sense. Explicitly mentioned and added an error message in ecd8e0f.

docs/releases.rst
psss: Wording updated in ecd8e0f.

docs/features.rst
LecrisUT: > > Probably should move these sections to be together in a fmf control section where all fmf metadata can be found. Currently `.fmf/version`, `.fmf/config`, `/:` configurations, etc.
> 
> Good, point. What about documenting `Trees` and `Config` close together? See [ecd8e0f](https://github.com/teemtee/fmf/commit/ecd8e0f0c90dd1c871b6b3f052206bcb235dabb1).

Was thinking more like in a future refactoring because some of these seem like advanced design tools. The restructuring I am thinking of is:
- "What is fmf": a simple explanation of fmf as yaml + file structure with inheritance.
- "Features": all the capabilities and definitions of fmf
- "Advanced designs": suggestions for how to structure the fmf tree, advanced in-tree control parameters like `/.inherit`, advanced tree control parameters in `.fmf/config`, full list of operators (just because can't find a better place)
- "Developer corner": what to expect from `.fmf/version`, how to expand fmf and more such topics

The concept page is too `tmt` specific and it's hard to parse outside of it. I think most of the content could be converted to a "what/why fmf".

---

But this is all for a different PR, I'll create an issue from this comment

LecrisUT (approved):

PR #265, opened 2024-12-03
Use `hatch version` instead of packit default

psss:
By default packit creates rpms with version which can be older than packages already released in Fedora repositories. We need to use `hatch version` to at least bump the micro version.

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -48,6 +48,7 @@ jobs:
     targets: *targets
     <<: *copr
     project: latest
+    release_suffix: "{PACKIT_PROJECT_BRANCH}"
 
   # Build release (copr stable)
   - job: copr_build
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -10,6 +10,8 @@ actions:
   create-archive:
     - "hatch build -t sdist"
     - "sh -c 'echo dist/fmf-*.tar.gz'"
+  get-current-version:
+    - hatch version
 
 # Common definitions
 _:
--- a/fmf.spec
+++ b/fmf.spec
@@ -15,6 +15,7 @@ BuildRequires: python3dist(docutils)
 BuildRequires: git-core
 Requires:      git-core
 
+Obsoletes:     python3-fmf < %{version}-%{release}
 %py_provides   python3-fmf
 
 %description


psss:
@LecrisUT, could you please have a look? I tried latest Fedora:

    > dnf install fmf
    > rpm -q fmf
    fmf-1.4.1-2.fc41.noarch
    > dnf copr enable @teemtee/latest
    > dnf update fmf
    Updating and loading repositories:
    Copr repo for latest owned by @teemtee             100% |   7.9 KiB/s |   1.5 KiB |  00m00s
    Repositories loaded.
    Nothing to do.

Here's what `packit` builds by default:

    fmf-1.4.1-1.20241203121310479530.main.9.g759788f

So, I guess we don't have another option.

LecrisUT:
Hmm, interesting, this is because packit puts commit details in `release` instead of `version`. I think it would be good to mimick `%forgemeta` approach, or the versioning guideline for snapshots in general. I'll raise the issue on packit

martinhoyer (approved):

LecrisUT (commented):
Until https://github.com/packit/packit/issues/2482 gets resolved, I think this would be a good workaround, but we would not be able to switch out of it until a next minor release because the predict next-version is bumping the patch version `1.4.2` and I am not sure how `.dev10` orders compared to `^20241203`.

Another approach would be:
```console
$ echo "$(git describe --abbrev=0 --tags)^$(date +%Y%m%d)"
```
Not sure how `sh -c` and escapes must be done for this though

psss:
I just realized we need to obsolete the dropped `python3-fmf` package otherwise it is not remove during updates. Should be covered in e7ee4e4.

psss:
> Until [packit/packit#2482](https://github.com/packit/packit/issues/2482) gets resolved, I think this would be a good workaround,

Why would a workaround be needed? The update path should work fine:

```
# starting on fedora
fmf-1.4.1-2.fc41

# update from the copr
fmf-1.4.2.dev11+g3ba9d95-1.20241203155123996628.h

# get the next stable when released
fmf-1.5.0-1.fc41
```

LecrisUT:
> > Until [packit/packit#2482](https://github.com/packit/packit/issues/2482) gets resolved, I think this would be a good workaround,
> 
> Why would a workaround be needed? The update path should work fine:
> 
> ```
> # starting on fedora
> fmf-1.4.1-2.fc41
> 
> # update from the copr
> fmf-1.4.2.dev11+g3ba9d95-1.20241203155123996628.h
> 
> # get the next stable when released
> fmf-1.5.0-1.fc41
> ```

The issue is when working on bug-fix versions like `fmf-1.4.2` which would clash with the `.dev` suffix that is added. It depends on how this part is ordered. The issue is once we use this, it would be hard to go cleanly got back to a packit default with snapshot syntax.

martinhoyer (approved):

psss:
> The issue is when working on bug-fix versions like `fmf-1.4.2` which would clash with the `.dev` suffix that is added. It depends on how this part is ordered.

Yeah, I guess you're right:

    > rpmdev-vercmp 1.4.2.dev11-1 1.4.2-1
    1.4.2.dev11-1 > 1.4.2-1

That was the reason for `tmt` hacks:

    > rpmdev-vercmp 1.40.dev9+gebf02add-main 1.40.0
    1.40.dev9+gebf02add-main < 1.40.0

So this means, that development bug-fix versions appear to be newer then the final. But as we'll have separate `stable` and `latest` repositories, I guess we're ok.

PR #269, opened 2024-12-04
Sync the `.packit.yaml` config to downstream repo

psss:
The config file needs to be present in the `rawhide` branch of dist-git to actually enable the `propose_downstream` feature. Let's keep the file in sync.

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -1,5 +1,7 @@
 specfile_path: fmf.spec
-files_to_sync: [fmf.spec]
+files_to_sync:
+  - fmf.spec
+  - .packit.yaml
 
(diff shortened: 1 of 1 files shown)


martinhoyer (approved):

LecrisUT (approved):
I don't remember if it's the spec file or the packit config that is synced by default, but it's good to have them both explicitly

psss:
I don't think there is any default value, [packit docs](https://packit.dev/docs/configuration#files_to_sync) state that only `packit init` would default for the two files:

> Spec file path and config file path are always included by packit init but can be manually removed from the list.

PR #271, opened 2025-01-06
Make sorting in `climb()` and `prune()` configurable

psss:
Allow users to disable sorting by node names and preserve the order in which child nodes where inserted into the tree.

## Checklist

* [x] implement the feature
* [x] write the documentation
* [x] mention the version
* [x] include a release note

### DIFF
--- a/docs/features.rst
+++ b/docs/features.rst
@@ -324,26 +324,6 @@ instead of deleting it completely. To do so, one can set the directive::
 By default all leaves have it set to ``true`` (such node is selected)
 and branches have set it to ``false`` (such node is not selected).
 
-.. _sort:
-
-Sort
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-
-By default, when exploring test metadata in the tree, child nodes
-are sorted alphabetically by node name. This applies to command
-line usage such as ``fmf ls`` or ``fmf show`` as well as for the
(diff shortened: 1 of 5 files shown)


lukaszachy (approved):

docs/features.rst
```diff
@@ -324,6 +324,26 @@ instead of deleting it completely. To do so, one can set the directive::
 By default all leaves have it set to ``true`` (such node is selected)
 and branches have set it to ``false`` (such node is not selected).
 
+.. _sort:
+
+Sort
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
LecrisUT, line 330: This document seems to be focused on the fmf file interface and currently this part is kinda out of place. Maybe we could add a `/.sort` (similar to `select`) interface. Otherwise I think having a python interface documentation would be better.

LecrisUT (commented):
When `sort=False` it is a bit confusing what order are child nodes created in w.r.t. scatter feature.

docs/features.rst
psss: Yeah, you're right that this new content is more related to Python coding. We already have the `Modules` section. Perhaps we can rename it to `Code` to have there all information related to Python code in a similar way as we have it for `tmt`. I've moved the new section there in c89892c. Will do the rename and cleanup in a separate pull request.

LecrisUT: I think it is more standard to call it `API Reference` or `Python (API) Reference`, but that's for another issue.

KwisatzHaderach (approved):

therazix (approved):

PR #272, opened 2025-01-06
Implement regular expression search for context

### LINKED ISSUE #226: Add operator for REGEX matching in context
Add two operators:
`~` for match
`!~` for not match

Note we have `~=` and similar for "major version" comparison.
However since `~` is used in tmt hardware it make sense to use such symbol here as well

psss:
Support new operator `~` for matching patterns and `!~` for negative matching patterns in the context rules.

Fix #226.

## Checklist

* [x] implement the feature
* [x] write the documentation
* [x] include a release note

### DIFF
--- a/fmf/context.py
+++ b/fmf/context.py
@@ -384,14 +384,14 @@ class Context:
         "~<=": _op_minor_less_or_eq,
         "==": _op_eq,
         "~=": _op_minor_eq,
-        "~": _op_match,
         "!=": _op_not_eq,
         "~!=": _op_minor_not_eq,
-        "!~": _op_not_match,
         ">=": _op_greater_or_equal,
         "~>=": _op_minor_greater_or_equal,
         ">": _op_greater,
         "~>": _op_minor_greater,
+        "~": _op_match,
+        "!~": _op_not_match,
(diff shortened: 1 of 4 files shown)


fmf/context.py
```diff
@@ -44,6 +44,9 @@ def __init__(self, origin):
         else:
             self._to_compare = self._split_to_version(origin)
 
+        # Store the original string for regexp processing
+        self.origin = origin
```
lukaszachy, line 48: Wouldn't `raw` be a better name for this variable? I find it more readable, especially when it is used in operators. Until now 'origin' was local to this init method so it didn't matter what was its name. (and yes, I'm bad at naming things :)

psss: Good idea, renamed in bd1ec88.

psss:
@lukaszachy, I also had to modify the operator matching a bit in 80b4b6e as `~` matched other operators which start with tilda. Please, have a look if it's ok. Thanks.

lukaszachy (approved):

KwisatzHaderach (approved):

therazix (approved):

PR #287, opened 2025-06-25
Pass the original node content to adjust decision callback

happz:
Because I would really like to show the diff between "before" and "after", and callback was getting just one version.

## Checklist

* [x] implement the feature

### DIFF
--- a/fmf/base.py
+++ b/fmf/base.py
@@ -39,16 +39,21 @@ class AdjustCallback(Protocol):
     A callback for per-rule notifications made by Tree.adjust()
 
     Function which will be called for every rule inspected by adjust().
-    It will be given three arguments: fmf tree being inspected,
-    current adjust rule, and whether the rule was skipped (``None``),
-    applied (``True``) or not applied (``False``).
+    It will be given the following arguments:
+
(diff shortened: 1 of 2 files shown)


fmf/base.py
```diff
@@ -584,16 +589,22 @@ def apply_rules(rule_set):
...
                         # Remove special keys (when, because...) from the rule
                         apply_rule = {
                             key: value
                             for key, value in rule.items()
                             if key not in ADJUST_CONTROL_KEYS
                             }
-                        self._merge_special(self.data, apply_rule)
+
+                        if decision_callback:
+                            node_before = self.copy()
```
LecrisUT, line 600: Should it `deep_copy` instead in case the merge changes something like a list, or is that already taken care of?

happz: I thought there's deepcopy used, but when I play with it I see no change between before and after, and I suspect the copy is not as deep as I expected...

happz: Nevermind, deepcopy is there, and what I saw was PEBKAC.

happz:
An example of tmt output when this becomes available - it borrows from the similar output of `--log-topic=policy`, note the "diff" of keys after a rule applies.

Issue #289, opened 2025-10-14
Allow `context` values to be special strings


LecrisUT:
The context of this is for `tmt`, but the capability should be enabled in `fmf` broadly.

The idea is that some context values we would want to override the comparator, e.g. for `distro` we should be more explicit what it means and why `distro < fedora-rawhide`. For these we can then use https://github.com/rpm-software-management/fedora-distro-aliases to support aliases like `fedora-development`.

On the `fmf` side what can be done is to expose a registry which will use a specific class constructor for the context dict instead of keeping it as a `str`.

LecrisUT:
Bumping the priority and considering it tentatively for next sprint because it is starting to affect Fedora ELN [^1] and the comparison there is quite ambiguous

[^1]: https://github.com/fedora-eln/eln/issues/471

junaruga:
I am looking forward to seeing this feature will be implemented.

PR #291, opened 2026-02-04
Add ContextValue version_cmp method tests for CentOS Stream and RHEL

junaruga:
Add the test_version_cmp_centos_stream and test_version_cmp_rhel.

The main purpose to add these tests is to make sure the `distro: centos-stream-rawhide` and `distro: rhel-X.rawhide` such as `distro: rhel-10.rawhide`, `distro: rhel-9.rawhide` work.
Because I heard that the `distro: rhel-10.rawhide` work

For the documentation, there is the following document about the `fedora-rawahide`. Let me know if you want to update the document with centos-stream and rhel.

https://fmf.readthedocs.io/en/stable/context.html#context
https://github.com/teemtee/fmf/blob/39af408279b305bd8744e1a74c4ad09958394587/docs/context.rst#major-version

> fedora-33 < fedora-rawhide ---> True (rawhide is newer than any number)

## Checklist

* [X] implement the feature
* [ ] write the documentation
* [X] extend the test coverage
* [ ] mention the version
* [ ] include a release note

### DIFF
--- a/tests/unit/test_context.py
+++ b/tests/unit/test_context.py
@@ -480,6 +480,40 @@ class TestContextValue:
         assert frawhide.version_cmp(f33) > 0
         assert f33.version_cmp(frawhide) < 0
 
+    def test_version_cmp_centos_stream(self):
+        """
+        CentOS Stream comparison
+        """
+
+        cs9 = ContextValue("centos-stream-9")
+        csrawhide = ContextValue("centos-stream-rawhide")
+
+        assert cs9.version_cmp(ContextValue("centos-stream-9")) == 0
+        assert cs9.version_cmp(ContextValue("centos-stream-8")) > 0
+        assert cs9.version_cmp(ContextValue("centos-stream-10")) < 0
+
+        assert csrawhide.version_cmp(ContextValue("centos-stream-rawhide")) == 0
+        assert csrawhide.version_cmp(cs9) > 0
+        assert cs9.version_cmp(csrawhide) < 0
+
+    def test_version_cmp_rhel(self):
+        """
+        RHEL comparison
+        """
+
+        rhel9_1 = ContextValue("rhel-9.1")
+        rhel9_rawhide = ContextValue("rhel-9.rawhide")
+
+        assert rhel9_1.version_cmp(ContextValue("rhel-9.1")) == 0
+        assert rhel9_1.version_cmp(ContextValue("rhel-9")) == 0
+        assert rhel9_1.version_cmp(ContextValue("rhel-9.0")) > 0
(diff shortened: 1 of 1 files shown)


LecrisUT:
This just feels wrong. My understanding is that the current comparison works just because of the accident of string comparison. I would prefer if we can start work on https://github.com/teemtee/fmf/issues/289, and instead define more explicitly how the context need to be ordered.

junaruga:
@LecrisUT Thanks for your response. I think you are right. The name `rhel-X.rawhide` means actually the latest minor version of the rhel-X, and `centos-stream-rawhide` means actually the latest version of the centos-stream. But there are no official definition of rawhide in RHEL and CentOS Stream. This may confuse people. So, I hope you guys would work on the feature #289.

> My understanding is that the current comparison works just because of the accident of string comparison. 

I suppose that the `ContextValue` `version_cmp` method calls `ContextValue.compare` method converting the string into integer at the following lines.

https://github.com/teemtee/fmf/blob/39af408279b305bd8744e1a74c4ad09958394587/fmf/context.py#L159-L161

However, while the comparing logic is to convert the version string to integer, there is already the document about the "fedora-33 < fedora-rawhide ---> True". And it seems the `rhel-X.rawhide` is already used in practice in my company.

junaruga:
It seems the CI testing-farm failures are not related to this PR.

For example, for the testing-farm:fedora-rawhide-x86_64 case:

https://artifacts.dev.testing-farm.io/8b1bb6af-012e-4798-87c2-58c1d092e346/

https://artifacts.dev.testing-farm.io/8b1bb6af-012e-4798-87c2-58c1d092e346/work-integrationf2bjxab0/plans/integration/execute/data/guest/default-0/tests/core/about-2/data/List-plugins-as-YAML/output.txt

```
:: [ 14:03:50 ] :: [   FAIL   ] :: Command 'yq '. | to_entries | [.[] | "\(.key): \(.value | sort | join(\" \"))"] | sort | .[]' /var/tmp/rlRun_LOG.laN6NHNJ > /tmp/tmp.O9vCfgKQ11/actual-plugin-list.txt' (Expected 0, got 3)
:: [ 14:03:50 ] :: [   FAIL   ] :: Command 'diff -u /tmp/tmp.O9vCfgKQ11/expected-plugin-list.txt /tmp/tmp.O9vCfgKQ11/actual-plugin-list.txt' (Expected 0, got 1)
```

LecrisUT:
Will move the handling to `tmt` since handling `fedora-eln` in the current framework is not possible (`rhel` resolves as higher than `fedora-eln` because of alphabetical check). First step towards this is in https://github.com/teemtee/fmf/pull/293

PR #293, opened 2026-04-28
Introduce `ContextDimension`

### LINKED ISSUE #289: Allow `context` values to be special strings

LecrisUT:
This is quite a big one, but I will keep the commits relatively logically contained for easier review commit by commit. The idea here is:
- Add a container for defining the context dimension comparison operators: `ContextDimension`
- Users can extend `ContextDimension` by sub-classing it and defining the `_dimension_name` it is linked to
- If a dimension is not found in the sub-classes, the `create_default` is used, which by default wraps around `ContextValue` for backwards-compatibility
- All of the comparator logic was moved to `ContextDimension` instead. `Operators` and decorators were added to manage these more dynamically
- Some cleanups
  - `staticmethod` -> `classmethod` to allow expanding the `Context` class
  - `re_expression_triple` was made a function for the reason above
  - Dropped some dead-code parts

Closes #289
Depends-on #297

---

## Checklist

* [x] implement the feature
* [ ] write the documentation
* [ ] mention the version
* [ ] include a release note

### DIFF
--- a/fmf/base.py
+++ b/fmf/base.py
@@ -495,7 +495,6 @@ class Tree:
             context,
             key='adjust',
             undecided='skip',
-            case_sensitive=True,
             decision_callback: Optional[AdjustCallback] = None,
             additional_rules=None,
             additional_rules_callback: Optional[ApplyRulesCallback] = None):
@@ -512,10 +511,6 @@ class Tree:
         context dimension is not defined. By default, such rules are
         skipped. In order to raise the fmf.context.CannotDecide
         exception in such cases use undecided='raise'.
-
-        Optional 'case_sensitive' parameter can be used to specify
-        if the context dimension values should be case-sensitive when
-        matching the rules. By default, values are case-sensitive.
 
         Optional 'decision_callback' callback would be called for every adjust
         rule inspected, with three arguments: current fmf node, current
@@ -558,8 +553,6 @@ class Tree:
             additional_rules = []
         elif isinstance(additional_rules, dict):
             additional_rules = [additional_rules]
-
-        context.case_sensitive = case_sensitive
 
         def apply_rules(rule_set):
             # 'continue' has to affect only its rule_set
@@ -624,7 +617,6 @@ class Tree:
         # Adjust all child nodes as well
         for child in self.children.values():
             child.adjust(context, key, undecided,
-                         case_sensitive=case_sensitive,
                          decision_callback=decision_callback,
                          additional_rules=additional_rules,
                          additional_rules_callback=additional_rules_callback)
--- a/fmf/base.py
+++ b/fmf/base.py
@@ -495,6 +495,7 @@ class Tree:
             context,
             key='adjust',
             undecided='skip',
+            case_sensitive: Optional[bool] = None,
             decision_callback: Optional[AdjustCallback] = None,
             additional_rules=None,
             additional_rules_callback: Optional[ApplyRulesCallback] = None):
@@ -532,6 +533,10 @@ class Tree:
         if not isinstance(context, fmf.context.Context):
             raise utils.GeneralError(
                 "Invalid adjust context: '{}'.".format(type(context).__name__))
+
+        # TODO: Remove this in next release
+        if case_sensitive is not None:
+            context._context_dimensions._default_dimension_cls.case_sensitive = case_sensitive
 
         # Adjust rules should be a dictionary or a list of dictionaries
         try:
@@ -617,6 +622,7 @@ class Tree:
         # Adjust all child nodes as well
         for child in self.children.values():
             child.adjust(context, key, undecided,
+                         case_sensitive=case_sensitive,
                          decision_callback=decision_callback,
                          additional_rules=additional_rules,
                          additional_rules_callback=additional_rules_callback)
--- a/fmf/context.py
+++ b/fmf/context.py
@@ -93,7 +93,7 @@ class ContextDimension(ABC, Generic[T]):
             _registrar = {}
 
         class TmtContext(fmf.context.Context):
-            _dimensions = TmtContextDimension
+            _context_dimensions = TmtContextDimension
 
         class DistroContextDimension(TmtContextDimension[DistroAlias]):
             _dimension_name = "distro"
@@ -512,7 +512,7 @@ class Context:
     """
     Represents https://fmf.readthedocs.io/en/latest/context.html
     """
-    # Operators' definitions
+    _context_dimensions: ClassVar[type[ContextDimension]] = ContextDimension
 
     def _op_defined(self, dimension_name, values):
         """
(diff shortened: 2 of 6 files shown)


tcornell-bus:
blocked by https://github.com/teemtee/fmf/pull/294

thrix:
This does not look good:

```
❯ python3 -c 'import fmf; fmf.Context("module = perl:5.28").matches("module == perl")'
  File "<string>", line 1, in <module>
    import fmf; fmf.Context("module = perl:5.28").matches("module == perl")
                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^
  File "/var/home/thrix/git/github.com/teemtee/fmf/fmf/context.py", line 752, in matches
    result = self.evaluate(expression)
  File "/var/home/thrix/git/github.com/teemtee/fmf/fmf/context.py", line 805, in evaluate
    return self._op_core(dimension_name, values, operator)
           ~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/var/home/thrix/git/github.com/teemtee/fmf/fmf/context.py", line 541, in _op_core
    assert isinstance(dimension_value, ContextDimension)
           ~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError
```

fmf/context.py
```diff
@@ -31,6 +38,292 @@ class InvalidContext(Exception):
...
+
+    @classmethod
+    def _make_value(cls, raw_value: str) -> "ContextValue":
+        return ContextValue(raw_value)
+
+    def _op_eq(self, other: str) -> bool:
+        return self.value.version_cmp(
+            self._make_value(other),
+            ordered=False,
+            case_sensitive=self.case_sensitive,
```
thrix, line 268: Case sensitivity is broken now

```
python - <<'PY'
from fmf.context import Context

context = Context(component="python3-3.8.5-5.fc32")
context.case_sensitive = False

print(context.matches("component == python3"))  # True
print(context.matches("component == PYTHON3"))  # Expected True, PR returns False
PY

True
False
```

fmf/context.py
```diff
@@ -31,6 +38,292 @@ class InvalidContext(Exception):
...
+    not subclassed and nothing is registered) the dimension values are treated as
+    :py:class:`ContextValue`.
+
+    .. code-block:: python
+
+        class TmtContextDimension(fmf.context.ContextDimension):
+            _registrar = {}
+
+        class TmtContext(fmf.context.Context):
+            _dimensions = TmtContextDimension
```
thrix: This should now be `_context_dimensions` ?

fmf/context.py
LecrisUT: Yes, the tests show that. Please look at the current design and provide a suggestion. There are various courses of action here:
- Delete the `case_sensitive` option and tests
- Fail on unsupported `case_sensitive` option
- Deprecate `case_sensitive` and do some ugly workaround
- Maintain the current support

There is a broader discussion necessary on the support, e.g. in tmt we are mixing both `case_sensitive = False` in various `adjust` and the default in `enabled_by_when` for example.

fmf/context.py
LecrisUT: Sure, but I will wait until there is a review on the design instead of ping-ponging between implementations. Please provide feedback on the design as well

fmf/context.py
thrix: I have no idea why the feature was added, I think we will need to find out first. But I would prefer first to provide a drop in replacement and work on dropping any functionality separately.

LecrisUT: Here are the facts I could gather:
- `tmt` is the only consumer of context other than `fmf` itself, and `fmf` cli does not expose context either. Let me know of instances that I did not look for, I considered `artemis-pools`, `rh-kernel-stqe/python-stqe`, and `fmf-jinja`
- `case_sensitive` was added in https://github.com/teemtee/fmf/pull/195, https://github.com/teemtee/tmt/pull/2357
- The case-sensitive handling in tmt is not consistent as seen in https://github.com/teemtee/tmt/blob/ee7b865b111c6f6d41114905431286ffd07ab807/tmt/steps/__init__.py#L2111-L2127

I personally don't see a reason to be conservative about this one since this change would need `tmt` change as well to be useful. There already is customizability and guidance if new users pop up. In the meantime, please review the design and the suggested way to opt-in the case-sensitivity! Right now the commits are very minimal and focused on the main functionality before it starts growing with the test adjustments and such.

@therazix @psss were involved in the introduction of the case-sensitive feature, so I will wait for their feedback and context in the meantime.

psss: Thanks much for preparing the draft of this. The overall design sounds good to me. There was a [tangible use case](https://github.com/teemtee/tmt/issues/1953) behind introducing the case-insensitive comparison so we definitely want to keep the insensitive handling at least. By default, I'd vote for keeping the option to choose the behavior. Defining the sensitivity at the `ContextDimension` class level looks like a good approach to me. But, if this would complicate the implementation too much, I wouldn't be strongly against dropping case-sensitive altogether, especially if there are no clear consumers for it.

Before diving into all related detailed changes, could we outline how the `is defined` and `is not defined` operators could be handled? Would it be some special separate value-less class? That way sounds a bit complicated to me. If possible, I'd rather suggest to keep a single `ContextDimension` and somehow allow value-less handling inside it. @LecrisUT, do you see a way how we could implement the second path?

LecrisUT: >  By default, I'd vote for keeping the option to choose the behavior. Defining the sensitivity at the `ContextDimension` class level looks like a good approach to me.

That should still be available. The part that I would prefer to be dropped is the interface at the `Context` level and make `ContextDimension` the source of truth for that instead. I could make the current `case_sensitive` api still work, but it would be ugly and have potential race condition-like bugs.

> Before diving into all related detailed changes, could we outline how the `is defined` and `is not defined` operators could be handled?

Good question, I do not have a good vision for those, here are some possibilities:
- Keep them at the `Context` level as special hard-coded values (current state)
- Extend the other 2 regex patterns with decorator and generalize them in `Context`
- ~~Same as :point_up: but try to add it to `ContextDimension` (somehow) instead~~ (quite difficult as I think about it, rather not)

> Would it be some special separate value-less class?

Don't think we would need that. If needed I think we can just split the operators decorator and add them to a separate registrar (options 2 and 3 in above). The main thing here would be to generalize the `Operators` so that it can hook to other class.

psss: > That should still be available. The part that I would prefer to be dropped is the interface at the `Context` level and make `ContextDimension` the source of truth for that instead. I could make the current `case_sensitive` api still work, but it would be ugly and have potential race condition-like bugs.

Sounds good to me to keep this at the `ContextDimension` level. Just to clarify on a tangible example: So if `tmt` would like to disable case sensitivity, it could inherit from `DefaultContextDimension` and change the class variable to `False` and handle all provided context dimensions with this custom class. No other change(s) needed?

> If needed I think we can just split the operators decorator and add them to a separate registrar (options 2 and 3 in above).

Separate register for unary and binary operators sounds ok to me.

LecrisUT: > So if `tmt` would like to disable case sensitivity, it could inherit from `DefaultContextDimension` and change the class variable to `False` and handle all provided context dimensions with this custom class. No other change(s) needed?

Exactly. I can update the tests to show this.

> > If needed I think we can just split the operators decorator and add them to a separate registrar (options 2 and 3 in above).
> 
> Separate register for unary and binary operators sounds ok to me.

:+1:

psss: > > So if `tmt` would like to disable case sensitivity, it could inherit from `DefaultContextDimension` and change the class variable to `False` and handle all provided context dimensions with this custom class. No other change(s) needed?
> 
> Exactly. I can update the tests to show this.

Good! :+1:

happz:
LGTM, in general the operator-based approach fits my view how these things should be implemented. I guess I will have a comment or two, mostly minor ones, when I get back in the evening.

fmf/context.py
```diff
@@ -31,6 +38,296 @@ class InvalidContext(Exception):
...
+    def _op_minor_greater_or_equal(self, other: str) -> bool:
+        raise NotImplementedError
+
+    @operators.add("~", "!~")
+    def _op_match(self, other: str) -> bool:
+        return re.search(other, self.raw_value) is not None
+
+
+@dataclass(frozen=True)
+class DefaultContextDimension(ContextDimension["ContextValue"]):
```
happz, line 239: What is the purpose of `DefaultContextDimension`? A short docstring would help.

LecrisUT: Added some more details for it

tests/unit/test_context.py
```diff
@@ -975,3 +978,66 @@ def test_minor_eq(self):
...
+                return self.value >= other
+
+            def _op_minor_eq(self, other: str) -> bool:
+                return self._op_eq(other)
+
+            def _op_minor_less(self, other: str) -> bool:
+                return self._op_less(other)
+
+            def _op_minor_less_or_equal(self, other: str) -> bool:
+                return self._op_minor_less_or_equal(other)
```
tcornell-bus:
```suggestion
                return self._op_less_or_equal(other)
```
gpt found a typo that could cause a recursion issue in the tests

fmf/context.py
tcornell-bus, line 1:
```mermaid

classDiagram
direction TD

class Tree {
  +adjust(context, key, undecided, case_sensitive, ...)
}

class Context {
  +_context_dimensions: type[ContextDimension]
  -_dimensions: dict[str, set[ContextDimension]]
  +parse_rule(rule)
  +split_expression(expression)
  +matches(rule)
  +evaluate(expression)
}

class Operators {
  +registrar: dict[str, tuple[str, bool]]
  +add(operator, negated_operator)
  +execute(operator, inst, other)
}

class ContextDimension~T~ {
  <<abstract>>
  +operators: Operators
  +raw_value: str
  +value: T
  +create(dimension_name, raw_value)
  +create_default(raw_value, dimension_name)
  +operate(operator, other)
  +_make_value(raw_value) T
  +_op_eq(other) bool
  +_op_less(other) bool
  +_op_less_or_equal(other) bool
  +_op_greater(other) bool
  +_op_greater_or_equal(other) bool
  +_op_minor_eq(other) bool
  +_op_minor_less(other) bool
  +_op_minor_less_or_equal(other) bool
  +_op_minor_greater(other) bool
  +_op_minor_greater_or_equal(other) bool
  +_op_match(other) bool
}

class DefaultContextDimension {
  +case_sensitive: bool <<class>>
  +dimension_name: str
  +_make_value(raw_value) ContextValue
}

class ContextValue {
  +raw: Any
  +version_cmp(other, minor_mode, ordered, case_sensitive) int
}

Tree --> Context : uses in adjust()
Context "1" *-- "*" ContextDimension : stores in _dimensions
Context --> ContextDimension : creates via _context_dimensions.create()
ContextDimension o-- Operators : operators registry
ContextDimension <|-- DefaultContextDimension
DefaultContextDimension --> ContextValue : wraps/comparisons

note for Tree "Temporary compatibility path:
adjust(case_sensitive=...) mutates
_default_dimension_cls.case_sensitive"
```

tcornell-bus: This helped me when my eyes started to hurt, maybe it'll help someone else. Putting in a thread to close it and get it out of the way if necessary.

fmf/context.py
thrix: Reviewing again

tcornell-bus (approved):

fmf/base.py
```diff
@@ -538,6 +534,10 @@ class describing the environment context. By default, the key
             raise utils.GeneralError(
                 "Invalid adjust context: '{}'.".format(type(context).__name__))
 
+        # TODO: Remove this in next release
+        if case_sensitive is not None:
+            context._context_dimensions._default_dimension_cls.case_sensitive = case_sensitive
+
```
thrix, line 540: Also mention the side effect of this, it mutates the classvar with all its consequences ...

thrix (approved):
Nothing else found. LGTM

fmf/base.py
LecrisUT: Does it make sense to be explicitly verbose about it when the comment above is stating remove this ASAP?

psss: Agree, the plan is to remove it soon, so I'd also say no need to document this more.

tests/unit/test_context.py
psss: Has this been addressed?

LecrisUT: Yeah, had it staged and was waiting to push it with other changes. Pushed it now 12c86ff9d83dee88aba712881c7f522b96ebad79

fmf/context.py
```diff
@@ -31,6 +40,306 @@ class InvalidContext(Exception):
...
+    def execute(self, operator: str, inst: "ContextDimension[T]", other: str) -> bool:
+        operator_func, negate = self.registrar[operator]
+        func: OperatorFunc = getattr(inst, operator_func)
+        if negate:
+            return not func(other)
+        return func(other)
+
+
+@dataclass(frozen=True)
+class ContextDimension(ABC, Generic[T]):
```
psss, line 81: Could you please add a short release note?

psss (approved):
Looks good. Just the missing release note.

fmf/context.py
LecrisUT: Decided to add a bare minimum release note 3c60ab5. Working on a proper documentation outside of this PR

psss: Thanks!

fmf/context.py
happz: Is there anything left to discuss in this thread?

therazix: > Is there anything left to discuss in this thread?

Should be informational only, closing.

PR #295, opened 2026-04-29
Update `/plans/integration` tests

LecrisUT:
The state of `/plans/integration` has been out-of-sync for a while, namely that we do not use python `yq` anymore. Ideally we would use the `plans/friends` in tmt to have tmt handle the necessary dependencies. That is waiting on https://github.com/teemtee/tmt/pull/4841.

In the meantime we are doing some cleanup of the plan here partially reflecting the design described in https://github.com/teemtee/tmt/pull/4841.
- Drop the usage of `fedora` branch which is not being updated automatically. There are some ideas of how we could bring it back to track which should be discussed in https://github.com/teemtee/tmt/pull/4841 and its followups
- Track the `@teemtee/latest` copr (tmt `main` branch)
- Add a design that is reusable in downstream packaging as well, basically taking the tmt files from the most relevant state of the dist-git

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -37,6 +37,8 @@ jobs:
   - job: copr_build
     trigger: pull_request
     targets: *targets
+    additional_repos:
+      - "copr://@teemtee/latest"
 
   # Test pull requests
   - job: tests
--- a/plans/integration.fmf
+++ b/plans/integration.fmf
@@ -2,7 +2,6 @@ summary:
     Run integration tests with tmt
 discover:
     how: fmf
-    url: https://github.com/teemtee/tmt
     filter: 'tier: 1 & tag:-provision-only'
 prepare:
   - how: install
@@ -10,3 +9,28 @@ prepare:
       - jq
       - yq
       - python3-pip
+enabled: false
+
+adjust:
+  - when: initiator == packit
+    because: >
+      We want to test against the tmt main branch
+    prepare+:
+      - how: install
+        copr: "@teemtee/latest"
+    discover+:
+      url: https://github.com/teemtee/tmt
+      ref: main
+    enabled: true
+  - when: initiator == fedora-ci
+    because: >
+      We want to test against the source in Fedora
+    discover+:
+      dist-git-source: true
+      # Technically we do not want to do dist-git-merge, but it is safe for us right now
+      # because there are no clashing test definitions
+      # https://github.com/teemtee/tmt/issues/4840
+      dist-git-merge: true
+      url: https://src.fedoraproject.org/rpms/tmt.git
+      ref: $@{dist-git-branch}
+    enabled: true
--- a/plans/integration.fmf
+++ b/plans/integration.fmf
@@ -3,23 +3,10 @@ summary:
 discover:
     how: fmf
     url: https://github.com/teemtee/tmt
-    ref: fedora
     filter: 'tier: 1 & tag:-provision-only'
 prepare:
   - how: install
     package:
       - jq
+      - yq
       - python3-pip
-      - tmt-all
-  - how: shell
-    script:
-      # Try to install into .local first, if that fails, try without --user.
-      # In the sane world, tmt would be running in a virtualenv, and we'd
-      # install yq there. Otherwise, we'd change system packages, but we'd
-      # have to run as root to do that, and who's running tmt test suite
-      # as root?
-      #
-      # We need to freeze yq to 3.1.1 so that it's installable on el8 as
-      # newer yq requires tomlkit>=0.11.7 which is python 3.7+ only.
-      - pip3 install --user yq==3.1.1 || pip3 install yq==3.1.1
-      - yq --help


plans/integration.fmf
```diff
@@ -3,23 +3,10 @@ summary:
 discover:
     how: fmf
     url: https://github.com/teemtee/tmt
-    ref: fedora
```
LecrisUT, line 6: @psss this branch has become quite dead, any objections for dropping it? We want to pick up the `@teemtee` copr repos in some way anyway. We had the same situation issue with `beakerlib` and `podman`, I will think about it a bit to see how to best resolve it.

.packit.yaml
```diff
@@ -37,6 +37,8 @@ jobs:
   - job: copr_build
     trigger: pull_request
     targets: *targets
+    additional_repos:
+      - "copr://@teemtee/latest"
```
LecrisUT, line 41: Technically we do not need it, but it's a fail-safe if there is any dependencies that tmt is picking up early before it lands in Fedora. It does not affect us negatively to add it at the very least

thrix:
@LecrisUT would you mind providing more details about the motivation here please? Description is too brief to understand what exactly you are addressing here and why. Would help with the review.

plans/integration.fmf
thrix: hmm, also the hotfix part of the release instructions still mentions it:

If we drop also the release docs needs an update, still mentions it as something we should

plans/integration.fmf
thrix, line 1: IF we are here, adding a description what is this plan used for would be nice.

LecrisUT:
> @LecrisUT would you mind providing more details about the motivation here please? Description is too brief to understand what exactly you are addressing here and why. Would help with the review.

Before even that, we need to go through the status of the repo settings. The description does not land in the commit details right now

plans/integration.fmf
LecrisUT: Suggestions please. I do not have anything more useful to add to the description that is not covered in the summary already

plans/integration.fmf
```diff
@@ -2,24 +2,35 @@ summary:
...
+adjust:
+  - when: initiator == packit
+    because: >
+      We want to test against the tmt main branch
+    prepare+:
+      - how: install
+        copr: "@teemtee/latest"
+    discover+:
+      url: https://github.com/teemtee/tmt
+      ref: main
```
thrix, line 23: I kinda would expect the integration plan is run against stable tmt, maybe if the description is clear where this plan is used, it would be easier to understand if this is what we want :)

LecrisUT: Not possible right now. Move that discussion in https://github.com/teemtee/tmt/pull/4841. Totally want make it available.

For fmf though, I don't think we should use stable. We should be picking up all the latest regressions as fast as possible.

plans/integration.fmf
LecrisUT: See description now

thrix: a lot better, ty

plans/integration.fmf
thrix: I would expect to see where is this integration plan planned to be used.

plans/integration.fmf
thrix: ack, seems there might be more use cases then

plans/integration.fmf
LecrisUT: I don't get it. The plan is running `tmt`'s tests. The plan is used in this repo's testing, just like any other plan in this repo: https://artifacts.dev.testing-farm.io/a635cf87-963f-4601-a598-e6a512c90d77/. There is no fancy packit filtering of plans or anything, it is just a regular tmt plan that is executed.

tcornell-bus (approved):
LGTM, tests passing in /plans/integration too

.packit.yaml
psss: Well, `fmf` does not (and should not) depend on `tmt` in any way. Why should we need to enable copr repo with latest `tmt` packages when building the `fmf` rpm package? What problems it should prevent?

plans/integration.fmf
```diff
@@ -2,24 +2,35 @@ summary:
...
+    enabled: true
+  - when: initiator == fedora-ci
```
psss, line 25: Currently we are not syncing plans from github to distgit so this change will have no effect. Do we want to enable the syncing as well. If so, maybe better to handle that in a separate pull request?

.packit.yaml
LecrisUT: As mentioned, building wise almost nothing. But it's `@teemtee/latest` not `tmt/latest` :wink:.

Let's consider this scenario (which applies to both `tmt` and `fmf`):
- downstream package lands an update to `python-ruamel-yaml` that broke compatibility
- they then bump the epoch and revert the update
- revert lands, but it is blocked by bodhi karma
- some downtime until copr can pick it up also

Using the `additional_repos` at PR stage allows us to add a temporary copr build (not necessarily of `tmt`) that we pick up until the dust settles, and minimize the downtime. It is effectively a no-op for the most parts, and mostly preventative for when we need it.

plans/integration.fmf
LecrisUT: Yep, plan is to enable that syncing. This here gives more context on why I chose the current design with the `adjust` and not went with linking to `https://github.com/teemtee/tmt` always. Either way, this part (+ manual sync to src.fp.o) or the PR enabling the sync would have to be blockers for fmf release.

psss: Ok, understood. Then let's keep this here as a preparation and enable the sync in a separate pull request.

.packit.yaml
psss: Ok, understood. Thanks for the example. On the other hand, could it also hide possible problems? Let's say we introduce a new feature in `fmf` which will not play nicely with the released `tmt` but would be working ok with the latest `tmt` as related changes would be already done there. This would prevent us from detecting the breakage before releasing a new `fmf`. Probably not very frequent scenario, but I remember cases when we broke some `tmt` use cases with a fresh `fmf`.

LecrisUT: > On the other hand, could it also hide possible problems? Let's say we introduce a new feature in `fmf` which will not play nicely with the released `tmt` but would be working ok with the latest `tmt` as related changes would be already done there.

Yes very much so. The only safe-guard for this that we have is the testing in downstream Fedora, and release/review process. PyPI is slightly more vulnerable, but at worst it would only be 2 weeks of incompatibility there. It should not affect tmt workflow at least.

> but I remember cases when we broke some `tmt` use cases with a fresh `fmf`.

`/plans/integration` is supposed to pick up the failures, but mirroring the dependency on tmt side would allow us to pick up the errors as well. This would require adding the `prepare` step to install the copr repo until https://github.com/packit/packit/issues/2719 lands (come to think of it, we can drop the copr install step here as well when that is resolved)

psss: Ok, let's turn this on for now.

psss (approved):
Agreed that the dist-git part of the integration plan should be followed by a pull request with enabling the plan syncing to downstream git repo.

plans/integration.fmf
thrix: ack, I kinda thought it is used from outside of this repo. I stand corrected.

PR #299, opened 2026-05-13
Fix symlink loop detection to allow shared symlink targets

psss:
The previous implementation used a single shared list of visited symlink targets across the entire tree. This caused sibling symlinks pointing to the same directory to be incorrectly skipped as loops, because the second symlink found its target already recorded by the first one.

Replace the shared list with a per-branch set of ancestor real paths. Each child node gets its own copy of the set, so sibling branches cannot interfere with each other. A symlink is now only skipped when its target resolves to an ancestor directory of the current walk path, which is the actual condition for a cycle.

Add tests for both the shared symlink target scenario and for the symlink loop detection to prevent future regressions.

Assisted-by: Claude Opus 4.6

## Checklist

* [x] implement the feature
* [x] include a release note

### DIFF
--- a/fmf/base.py
+++ b/fmf/base.py
@@ -104,12 +104,11 @@ class Tree:
         # Special directives
         self._directives = dict()
 
-        # Store symlinks in while walking tree in grow() to detect
-        # symlink loops
+        # Track real paths of ancestor directories to detect symlink loops
         if parent is None:
-            self._symlinkdirs = []
+            self._realpaths = set()
         else:
-            self._symlinkdirs = parent._symlinkdirs
+            self._realpaths = set(parent._realpaths)
 
         # Special handling for top parent
         if self.parent is None:
@@ -708,6 +707,7 @@ class Tree:
         if path in IGNORED_DIRECTORIES:  # pragma: no cover
             log.debug("Ignoring '{0}' (special directory).".format(path))
             return
+        self._realpaths.add(os.path.realpath(path))
         log.info("Walking through directory {0}".format(
             os.path.abspath(path)))
         try:
@@ -754,16 +754,10 @@ class Tree:
                 continue
             fulldir = os.path.join(dirpath, dirname)
             if os.path.islink(fulldir):
-                # According to the documentation, calling os.path.realpath
-                # with strict = True will raise OSError if a symlink loop
-                # is encountered. But it does not do that with a loop with
-                # more than one node
                 fullpath = os.path.realpath(fulldir)
-                if fullpath in self._symlinkdirs:
+                if fullpath in self._realpaths:
                     log.debug("Not entering symlink loop {}".format(fulldir))
                     continue
-                else:
-                    self._symlinkdirs.append(fullpath)
 
             # Ignore metadata subtrees
             if os.path.isdir(os.path.join(path, dirname, SUFFIX)):
--- a/tests/unit/test_base.py
+++ b/tests/unit/test_base.py
@@ -575,6 +575,59 @@ class TestTree:
         with pytest.raises(fmf.utils.JsonSchemaError):
             self.wget.find('/recursion/deep').validate('invalid')
 
+    def test_symlink_shared_target(self):
+        """
+        Multiple symlinks pointing to the same directory should all be followed
+        """
+
+        directory = tempfile.mkdtemp()
+        try:
+            Tree.init(directory)
+            common = os.path.join(directory, 'common')
+            os.mkdir(common)
+            with open(os.path.join(common, 'main.fmf'), 'w') as main:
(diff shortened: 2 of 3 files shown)


psss:
@petr-janda, this should fix the problem with directory symlinks.

tests/unit/test_base.py
```diff
@@ -575,6 +575,59 @@ def test_validation_invalid_schema(self):
...
+    def test_symlink_shared_target(self):
+        """
+        Multiple symlinks pointing to the same directory should all be followed
+        """
+
+        directory = tempfile.mkdtemp()
+        try:
+            Tree.init(directory)
+            common = os.path.join(directory, 'common')
+            os.mkdir(common)
```
LecrisUT, line 587: Please let's use `pathlib`

tests/unit/test_base.py
```diff
@@ -575,6 +575,59 @@ def test_validation_invalid_schema(self):
         with pytest.raises(fmf.utils.JsonSchemaError):
             self.wget.find('/recursion/deep').validate('invalid')
 
+    def test_symlink_shared_target(self):
+        """
+        Multiple symlinks pointing to the same directory should all be followed
+        """
```
LecrisUT, line 581: Could you embed the output of `tree` in here. If I follow this correctly it is:
```console
$ tree -al .
.
├── common
│   └── main.fmf
├── .fmf
│   └── version
├── one
│   ├── main.fmf
│   └── plan -> ../common  [recursive, not followed]
└── two
    ├── main.fmf
    └── plan -> ../common  [recursive, not followed]
```

fmf/base.py
```diff
@@ -756,16 +756,10 @@ def grow(self, path):
...
-                # with strict = True will raise OSError if a symlink loop
-                # is encountered. But it does not do that with a loop with
-                # more than one node
                 fullpath = os.path.realpath(fulldir)
-                if fullpath in self._symlinkdirs:
+                if fullpath in self._realpaths:
                     log.debug("Not entering symlink loop {}".format(fulldir))
                     continue
-                else:
-                    self._symlinkdirs.append(fullpath)
```
LecrisUT, line 766: Counterpoint, why do we not let python handle this? For example:
```console
$ tree -al .
.
├── common
│   ├── main.fmf
│   └── sub_plan
│       └── main.fmf
├── .fmf
│   └── version
├── one
│   ├── main.fmf
│   └── plan -> ../common  [recursive, not followed]
└── two
    ├── main.fmf
    └── plan -> ../common  [recursive, not followed]
```
```python
>>> from pathlib import Path
... for root, dirs, files in Path(".").walk(follow_symlinks=True):
...     print(f"root={root}, dirs={dirs}, files={files}")
root=., dirs=['common', 'one', 'two', '.fmf'], files=[]
root=common, dirs=['sub_plan'], files=['main.fmf']
root=common/sub_plan, dirs=[], files=['main.fmf']
root=one, dirs=['plan'], files=['main.fmf']
root=one/plan, dirs=['sub_plan'], files=['main.fmf']
root=one/plan/sub_plan, dirs=[], files=['main.fmf']
root=two, dirs=['plan'], files=['main.fmf']
root=two/plan, dirs=['sub_plan'], files=['main.fmf']
root=two/plan/sub_plan, dirs=[], files=['main.fmf']
root=.fmf, dirs=[], files=['version']
```
I remembered I had a snippet for how to manipulate the `walk()`, I'll try to dig it back up

LecrisUT: Ah, yes, `os.walk` does that trick, here's an example of skipping going down `common`
```python
>>> for root, dirs, files in os.walk(".",followlinks=True):
...     if root == ".":
...         dirs.remove("common")
...     print(f"root={root}, dirs={dirs}, files={files}")
...     
root=., dirs=['one', 'two', '.fmf'], files=[]
root=./one, dirs=['plan'], files=['main.fmf']
root=./one/plan, dirs=['sub_plan'], files=['main.fmf']
root=./one/plan/sub_plan, dirs=[], files=['main.fmf']
root=./two, dirs=['plan'], files=['main.fmf']
root=./two/plan, dirs=['sub_plan'], files=['main.fmf']
root=./two/plan/sub_plan, dirs=[], files=['main.fmf']
root=./.fmf, dirs=[], files=['version']
```

LecrisUT: Ah I tried it on `test_symlink_loop` and it misbehaved by default. Can still use the tricks with `os.walk` to simplify things, but it doesn't work by default, can postpone this.

petr-janda:
> @petr-janda, this should fix the problem with directory symlinks.

Yes it does. With this patch I can symlink a directory with plans to multiple directories. Together with inheritance it eases my life. Thank you.

psss:
Great, thanks for verifying!

PR #300, opened 2026-05-15
Align Context to a read-only dict interface

LecrisUT:
This aligns it with https://github.com/teemtee/tmt/pull/4892 so that they can be later unified and avoid translating between the two after #293


happz (approved):

psss (approved):

psss:
Some unit tests are failing.

PR #303, opened 2026-08-07
Implement plugin/modular system for FMF metadata loaders (Phase 1)

### LINKED ISSUE #103: Support exploring metadata from Python source code
@jscotka has experimented with storing `fmf` metadata in python decorators:

 * https://github.com/jscotka/fmf_metadata

It could be useful to add support for such parsing directly to `fmf`. By default, `*.fmf` files would be checked only, a dedicated option could be used to enabled parsing `*.py` files as well.

    fmf --extension py

Metadata could be combined from both sources:

    fmf --extension fmf --extension py

In `tmt` the discover step could support a relevant option to enable such behaviour:

    discover:
        how: fmf
        extension: [fmf, py]

jscotka:
`desciption changed, to be aligned with current implementation`

## Plugin/Modular System for FMF Metadata Loaders (Phase 1)

Adds a plugin architecture so fmf can read metadata from formats beyond `.fmf`
(YAML). This is the infrastructure phase; `.fmf` loading is refactored into the
first built-in plugin and stays the default. Backward compatible — `.fmf`-only
trees behave exactly as before.

Closes #103

### What it implements

- **`Plugin` base class** (`fmf/plugin.py`) — `read()`, `write()`,
  `read_config()`, and a default `can_handle()`.
- **`PluginRegistry`** (`fmf/plugin_loader.py`) — discovers plugins and
  dispatches each file to the first one that can handle it.
- **`FmfPlugin`** (`fmf/plugins/fmf.py`) — existing `.fmf`/YAML loading as a
  plugin (round-trip safe).
- **`Tree` integration** — loading and the experimental write-back
  (`__enter__`/`__exit__`) go through the registry.

### Key design points

- **Entry-point registration** — plugins advertise under the `fmf.plugins`
  group; no hardcoded list. Installed plugins are picked up automatically.
- **Fixed load order** — first plugin that can handle a file wins, in module
  discovery order. Not user-configurable (no priority setting).
- **Per-plugin config** — each plugin reads its own section of `.fmf/config`
  via `config_section` + `read_config()` (applied once per tree).
- **Fast on large trees** — plugin instances and compiled file patterns are
  cached; config is applied once per tree, not per file. `.fmf` parsing speed
  is unchanged.
- **Format version stays `1.x`** — the change is additive.

### Docs & roadmap

`docs/concept.rst` documents the plugin system; `PLUGIN_FUTURE.md` covers the
roadmap — Phase 2 (bash), Phase 3 (pytest, AST-only), Phase 4 (write-back).


docs/concept.rst
```diff
@@ -123,6 +123,91 @@ In the example above files or directories named ``.plans`` or
 the ``.fmf`` directory cannot be used for storing metadata.
 
 
+Plugins
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Starting with fmf 2.0, a plugin system allows reading metadata from
```
LecrisUT, line 129: Why 2.0? That would need a whole lot of other discussion rounds and gathering cleanups to do. For now it is additive so can skip that.

Also there are proper sphinx directives to indicate which version it targets

docs/concept.rst
```diff
@@ -123,6 +123,91 @@ In the example above files or directories named ``.plans`` or
...
+Plugin Priority
+---------------
+
+When multiple plugins can handle the same file extension, the plugin
+with the highest priority (0-200) is selected. Built-in plugins use
+these priorities:
+
+* FmfPlugin: 100 (default format)
+* BashPlugin: 50 (future)
+* PytestPlugin: 50 (future)
```
LecrisUT, line 169: Why is the priority even needed?

fmf/plugins/__init__.py
```diff
@@ -0,0 +1,23 @@
...
+"""
+
+from fmf.plugin_loader import get_registry
+# Import and register all built-in plugins
+from fmf.plugins.fmf import FmfPlugin  # noqa: F401
+
+# Register built-in plugins
+_registry = get_registry()
+_registry.register(FmfPlugin)
+
```
LecrisUT, line 14: With a new implementation it should use modern designs such as using `entry-point`. When starting from scratch here would not even bother providing many ways to define these, not even a hard-coded initial dict

fmf/plugins/fmf.py
```diff
@@ -0,0 +1,109 @@
...
+            filename: Path to .fmf file
+
+        Returns:
+            Dictionary with metadata, or empty dict if file is empty
+
+        Raises:
+            FileError: If file cannot be parsed or contains duplicate keys
+        """
+        try:
+            with open(filename, encoding='utf-8') as datafile:
```
LecrisUT, line 67: Use modern `pathlib`

LecrisUT (commented):
Some initial design comments

docs/concept.rst
jscotka: I've also though about that. and for me it also seems good idea to increase version, as this is significant change and probably we should also change version of `.fmf/version` to `2.0`
But theoretically it does not matter so much, as it is backward compatible. So yes, we can remove this part at all if you want.

docs/concept.rst
jscotka: For me it seems more robust and flexibile, as FMF should have higest prio when loading data, and bash and python files as well. I've expected that someone can make it overlapping. so that this priority can make plugin handling more flexible. e.g. imagine situation, that someone decide to rewrite some testsuite from bash to python, and will have old bash code here with new tests (with same tests and same naming) causes that if bash will have higher prio by default, override python metadata. 

I understand that this is probably corner case, and we can/couldn't support this scenario at all.
Discussion is welcomed. I can remove also this part as not crucial. Theoretically we can also add it later on demand

fmf/plugins/__init__.py
jscotka: I'll rework it. I've never used this `entry-point` so didn't know it exists :-)

fmf/plugins/fmf.py
jscotka: yes I can change it, it is older relict from old past implementation.

docs/concept.rst
LecrisUT: Iiuc the only difference is if you have both an fmf and pytest/bash node, which order does the equivalent of [elasticity](https://fmf.readthedocs.io/en/stable/features.html#elasticity) feature resolve? We do not have it well-defined there either :/

But I would rather we well-define the order and let the rest be a user issue until we get some context on how a user would prefer. I do not have strong preference if fmf files should be on top of pytest/bash, I would mostly use it as additive to each other. The only thing that comes to mind are the `/:` or `adjust` features, which should be in fmf files only IMO, and for that probably fmf definition being on the bottom would be needed?

jscotka: yes, this is completely right, that it is not well defined as well now, e.g. use `main.fmf` with key `/a/b/c` and then `/a/b/c/main.fmf` so that they are definig same item on same level twice. We've tried to solve in past as I remember some discussion but without success, or better to say no decision :-)

Yes, We can do some sort of harcoded priority, probably better than this flexibility.

For FMF files It is little bit harder, there are two points, first is that `main.fmf` will have still lowest prio, as this creates hierarchy (I do not expect to use some `main.sh`, or `main.py` to be used for same thing) but still as the most flexible and cleanest format it should be able to override/append data into another types. It is connected with my idea that if some plugin/format will not support writing data back to the source code (hard to do or create crapp in source code), the easiest way is to overrride them with  fmf file `+` items, so that most of data will live with tests and then e.g. generated IDs could be part or fmf files for same level of data.

jscotka: I've forgot to mention also main idea about that priority, what was in my mind :-) 

I plan to write two python plugins, one will support using just *doc strings* (for project what do not want to be dependent on `FMF` project using annotaded yaml strip inside) and second one with decorators like some  `@FMF.author("A B")`  -- and in this case  I cannot decide situation when somebody wants to use both together. 

@cle what do you think in this case. also hardcode the prio based on my decision? Or  allow to use both ways  inside one plugin together, so that, decision and documentation will be then on this plugin itself (but this part could cause some nondeterministic behaviour where to write data back into files, theoretically could be also defined inside documentation for plugin in case it will be part of one plugin)

LecrisUT: > as this creates hierarchy (I do not expect to use some main.sh, or main.py to be used for same thing)

I do, for example

```python
# test_someting.py

"""
.. fmf::
    duration: 10m
"""

def normal_test(): ...

def long_test(): ...
    """
    .. fmf::
        duration+: "* 10"
    """
```

> @cle what do you think in this case. also hardcode the prio based on my decision? Or allow to use both ways inside one plugin together, so that, decision and documentation will be then on this plugin itself (but this part could cause some nondeterministic behaviour where to write data back into files, theoretically could be also defined inside documentation for plugin in case it will be part of one plugin)

I would say they should be mutually exclusive.

As for the order itself, if we do not resolve the order issue in fmf files itself, I think we can assume for now that the user will play nice and not override willy-nilly. Or at least make it bright red in the documentation that this is a yet to be defined behavior.

PS: poor `@cle` being pingedout of the blue.

docs/concept.rst
LecrisUT: Agreed on last hacking session to keep 1.x for now

docs/concept.rst
LecrisUT: Agreed on last hacking session to not have it user-defined. There were no strong opinions if fmf should be on top of python or vice-versa, so fine to make a case for either one. Similar with the `elasticity` we do  not have a well-documented behavior and just hope the user doesn't override it, but we should document what the current method is for `elasticity` and similarly whatever we go for here.

fmf/plugins/__init__.py
LecrisUT: From hacking session, only one method and `entry-point` can cover some more complex edge-cases, so let's go with it.

jscotka:
Hi @LecrisUT,

I've fixed several issue what you've pointed in PR review, also I've tried to fix some issue what was in my mind when changing stuff -> what came to my mind was mainly performance issue with plugin system. So I've also tried to implement it in better way.

PLUGIN_FUTURE.md
```diff
@@ -0,0 +1,553 @@
+# FMF Plugin System - Future Development Roadmap
```
LecrisUT, line 1: Could you avoid committing this file? We don't use roadmap-like file files or documents here

docs/concept.rst
```diff
@@ -123,6 +123,107 @@ In the example above files or directories named ``.plans`` or
...
+When more than one plugin can handle the same file, the first plugin in
+load order (the order in which the plugin modules are discovered) is used.
+The order is not user configurable.
+
+.. warning::
+
+    As with :ref:`elasticity <elasticity>`, the resolution order when the
+    *same* node is defined by several formats (for example a ``.fmf`` file
+    and a source file at the same level) is not yet fully specified. Avoid
+    relying on cross-format overrides until the behavior is well defined.
```
LecrisUT, line 164: Would put these into a dedicated section for developers. Everything else in here is much more simplified than this and it is more user orientated. For now can put it in `contribute` until we figure a better structure.

docs/concept.rst
```diff
@@ -123,6 +123,107 @@ In the example above files or directories named ``.plans`` or
...
+    [project.entry-points."fmf.plugins"]
+    bash = "mypackage.plugins:BashPlugin"
+
+Any plugin installed in the current environment is picked up
+automatically.
+
+Loading order
+-------------
+
+When more than one plugin can handle the same file, the first plugin in
```
LecrisUT, line 155: handle the same file -> node. But this could also be a note/warning in `elasticity` section that is expanded for plugins also.

docs/concept.rst
```diff
@@ -123,6 +123,107 @@ In the example above files or directories named ``.plans`` or
...
+            self.settings = self.config_section_data(config)
+
+The ``config_section_data()`` helper returns the mapping stored under the
+plugin's ``config_section`` (or an empty dict when the section is missing
+or the plugin defines no section). The built-in ``FmfPlugin`` reads the
+``fmf`` section; it currently has no options and the section is reserved
+for future settings.
+
+Because the config is identical for every node in a tree, ``read_config()``
+is called only *once per tree*, not for each file.
```
LecrisUT, line 199: This is mixing user and developer info. The user is only interested how these are enabled or configured. The rest is developer focused and should have its dedicated sections

docs/concept.rst
```diff
@@ -123,6 +123,107 @@ In the example above files or directories named ``.plans`` or
...
+not free (``FmfPlugin`` builds a YAML parser in its constructor), so caching
+the instance avoids a per-file cost that would otherwise dominate on large
+trees. Plugins must therefore keep no per-file state: :meth:`read` and
+:meth:`write` receive the filename as an argument, and configuration is
+applied once via ``read_config()``.
+
+File matching is also kept cheap. The default ``can_handle()`` checks the
+plugin's ``extensions`` (a plain suffix test) first and only then its
+``file_patterns``, which are compiled to regular expressions once and cached
+on the class rather than recompiled for every file.
```
LecrisUT, line 216: I don't think this is valuable

docs/concept.rst
```diff
@@ -123,6 +123,107 @@ In the example above files or directories named ``.plans`` or
...
+
+Available plugins
+-----------------
+
+* ``fmf.plugins.fmf.FmfPlugin`` -- YAML-based ``.fmf`` files (built-in)
+
+Additional plugins (Bash scripts, Python/pytest tests) are planned; see
+``PLUGIN_FUTURE.md`` for the implementation roadmap.
```
LecrisUT, line 224: It should be a dedicated page, and drop the mention to not yet developed plugins or files that are not rendered or navigable to.

fmf/plugins/__init__.py
```diff
@@ -0,0 +1,8 @@
+"""
+FMF Plugins Package
+
+Built-in plugins live in this package and are advertised to the plugin
+registry through the ``fmf.plugins`` entry point group (see ``pyproject.toml``).
+They are discovered automatically at runtime; there is no hard-coded plugin
+list to maintain here.
+"""
```
LecrisUT, line 8:
```suggestion
"""
fmf plugins
"""
```
- `fmf` is the canonical naming schema
- Negative information of what is not done is not useful
- We don't need much information, at most point it to the rst document
- Do not duplicate information, e.g. where entry points are defined. I am fine with it being either in this docstring or a dedicated page, but not both.

fmf/plugins/fmf.py
```diff
@@ -0,0 +1,107 @@
+"""
+FMF Plugin - YAML-based metadata loader
+"""
```
LecrisUT, line 3:
```suggestion
"""
fmf plugin handling ``.fmf`` files
"""
```

fmf/plugins/fmf.py
```diff
@@ -0,0 +1,107 @@
...
+from typing import Any, Dict, List
+
+from ruamel.yaml import YAML
+from ruamel.yaml.error import YAMLError
+
+import fmf.utils as utils
+from fmf.plugin import Plugin
+from fmf.utils import dict_to_yaml, log
+
+# Constants for .fmf files (moved from base.py)
```
LecrisUT, line 15:
```suggestion
```
Drop, it is self-evident ("constants for ...") or meaningless to the current state ("moved from ...")

fmf/plugin.py
```diff
@@ -0,0 +1,230 @@
...
+          nor ``file_patterns`` can express
+        - write(): Method to write metadata back (default: fallback to .fmf)
+        - config_section: Name of the plugin's section in ``.fmf/config``
+    """
+
+    # Class attributes to be defined by subclasses
+    extensions: List[str] = []  # File extensions, e.g., [".fmf"]
+    file_patterns: List[str] = []  # Regex patterns for filenames
+    # Name of this plugin's section in the .fmf/config file (None = no section)
+    config_section: Optional[str] = None
```
LecrisUT, line 35: Class attributes are documentable with either `#:` or `"""`. We are using the former.

fmf/plugin.py
```diff
@@ -0,0 +1,230 @@
...
+        - file_patterns: List of regex patterns to match filenames
+        - can_handle(): Override only for matching that neither ``extensions``
+          nor ``file_patterns`` can express
+        - write(): Method to write metadata back (default: fallback to .fmf)
+        - config_section: Name of the plugin's section in ``.fmf/config``
+    """
+
+    # Class attributes to be defined by subclasses
+    extensions: List[str] = []  # File extensions, e.g., [".fmf"]
+    file_patterns: List[str] = []  # Regex patterns for filenames
```
LecrisUT, line 33: My vote is to drop these, let the plugin figure out how to define how to scan and get its own files. E.g. for `pytest` we could either find the `.py` files ourselves or let `pytest` discover them including dynamic ones. I can see value in both and it is a topic for a later date, but for nowI would not confine it.

fmf/plugin.py
LecrisUT, line 35:
```suggestion
    CONFIG_SECTION: ClassVar[str]
```
- naming schema for constants or class variables
- do not instantiate. It keeps the typing more constrained. I remember that typing can enforce type-checkers to error if not defined

fmf/plugin.py
```diff
@@ -0,0 +1,230 @@
...
+        is called for every file in a tree, so compiling on each call (or
+        relying on the standard library's evictable regex cache) would add an
+        avoidable per-file cost on large trees.
+        """
+        # Use the class' own __dict__ so subclasses do not reuse a parent cache
+        cache = cls.__dict__.get("_file_patterns_re")
+        if cache is None:
+            cache = [re.compile(pattern) for pattern in cls.file_patterns]
+            cls._file_patterns_re = cache  # type: ignore[attr-defined]
+        return cache
```
LecrisUT, line 52: Super ugly, drop, and any of its usage.

fmf/plugin.py
```diff
@@ -0,0 +1,230 @@
...
+
+        Args:
+            filename: Absolute or relative path to file
+
+        Returns:
+            True if plugin can read this file, False otherwise
+        """
+        if any(filename.endswith(extension) for extension in self.extensions):
+            return True
+        return any(pattern.search(filename) for pattern in self._compiled_patterns())
```
LecrisUT, line 70: As mentioned before would rather have the plugins figure everything related to this themselves

fmf/plugin_loader.py
```diff
@@ -0,0 +1,148 @@
...
+
+
+def _iter_entry_points(group: str) -> Iterable:
+    """Return entry points for the given group (Python 3.9+ compatible)."""
+    discovered = entry_points()
+    # Python 3.10+ returns an EntryPoints object with select(),
+    # Python 3.9 returns a plain dict keyed by group name.
+    if hasattr(discovered, "select"):
+        return discovered.select(group=group)
+    return discovered.get(group, [])  # type: ignore[attr-defined]  # Python 3.9
```
LecrisUT, line 21: We are using `_compat` pattern and we import a newer `importlib_metadata` entry_points` when needed.

fmf/plugin_loader.py
```diff
@@ -0,0 +1,148 @@
...
+
+
+class PluginRegistry:
```
LecrisUT, line 24: I do not like this pattern. This should be part of `Plugin` class itself, i.e. the registry is inside the `Plugin` and uses `classmethod`s. See `__init_subclass__`

LecrisUT (changes requested):
Quick initial comments because I got tired midway. Will come again when I get more energy to go through this

PLUGIN_FUTURE.md
jscotka: yes, I'll remove it, and I'll let it just locally. From the perspective of review I've commited and pushed it as well, to be clean what will be next steps for this PR to understand the architecture and why I've done something there.

LecrisUT: Too much text will have the opposite effect though. Do keep that in mind. Would much prefer a short one sentence in the description, or a comment when (or rather _if_) it becomes relevant

